PR #1, opened 2019-03-14
Re-convert 2016 LaTeX

davidfarmer:
The files in ptxConversion0 were made by converting CalculusI.tex from July 2016.

A diff between those and the V5 converted files will give some clues for how
to update the working copy of the PreTeXt files.

### DIFF
--- a/ptxConversion0/bibliography.ptx
+++ b/ptxConversion0/bibliography.ptx
@@ -1,0 +1,6 @@
+
+
+<references>
+  <title>Bibliography</title>
+</references>
+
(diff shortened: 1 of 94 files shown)


sean-fitzpatrick:
Since this seems to be just adding a new folder, I see no reason not to merge it.

sean-fitzpatrick:
So, this is an unprocessed conversion of APEX V3, correct?
The point of this I suppose is then to see what changes were made by @APEXCalculus between V3 and V5. Those changes could then be merged into the processed source at PCCMath's apex-mbx.

That seems reasonable to me.

PR #3, opened 2019-06-10
Update Section 1.1

sean-fitzpatrick:
Decided to have a go at the first section to see how it goes.
I did a 3-way diff using Meld, with this file in the middle; on the left, the apex-mbx version, and on the right, the ptxconversion0 version.
The one on the right let me catch a few small edits and typo fixes made by Greg after V3.
The one on the left let me pull in the TikZ code, WeBWorK exercises, etc.

Most of the other changes coming from apex-mbx are:
- adding <m></m> wrappers around numbers in text
- adding \text{.} in display math to get the period in the right place
- labelling tables as such (rather than as figures)
- a few LaTeX improvements, like using the cases environment, and getting rid of \displaystyle calls.
(I did change one `\ds\frac{}{}` to `dfrac{}{}`. Not sure what the PreTeXt rules are around use of \dfrac, but I found one \dfrac in the Sample Article so I figured it must be OK.)


sean-fitzpatrick:
Ha. Didn't realize a second commit would get added to the same pull request.
Anyway, I also removed a bunch of unnecessary files, like pdfs.

But I should have kept the /text folder in case @davidfarmer wants to run a further conversion off of this repo. (The alternative is to just use APEX V5 for this.)

So maybe don't approve the PR yet, or ask me to rewind the last commit.

APEXCalculus:
I'll let David decide on this pull request. If he has no use of the /text folder, he can go ahead and merge. If he does have use for it, you can roll it back.

It seems to me that he would want it for later sections that didn't get converted last time. I.e., the current ptx folder has only sections going through chapter 6. So that leaves a lot left to convert with David's program. But yes, he can get those files from the APEXCalcV5 repo.

davidfarmer:
In PreTeXt the period at the end of display math should be after the closing me or md tag.
(The LaTeX produced from the PTX will have the correct markup.)

I am going to merge and then use APEX V5 for the missing files.

sean-fitzpatrick:
Thanks. In the PTX code from PCCMathSAC/apex-mbx, periods and commas within display math are handled as `<me>some math \text{,}</me>`
Machine conversion gives `<me>some math</me>,`

From the last comment I understand the latter to be correct -- perhaps a change in PreTeXt since the last go around. I'll account for this in the next round of edits.

davidfarmer:
Handling the correct markup (bare punctuation outside the "me")
is relatively recent for both PreTeXt and my conversion script.

davidfarmer:
Me being pedantic:

It is my prettyprinter, not the conversion script, that rearranges
the punctuation.

PR #5, opened 2019-06-12
Merge remaining edits for Chapter 1

sean-fitzpatrick:
Finished with the remaining sections for Chapter 1.
I also updated the chapter file, and index.ptx.
(The latter is a copy of the apex-mbx one.)

For these sections it was mostly a matter of identifying the changes made by Greg moving from V3 to V4 and incorporating them into the previous conversion attempt.

There were a few new problems added. I coded these as WeBWorK problems, and I think most of them are OK. (One I'm not sure about, in the continuity section.)

Found one instance of "behaviour" in the apex-mbx version. Was patriotically tempted to keep it, but I dropped the u.


sean-fitzpatrick:
I made a few minor changes to account for suggestions from Greg and Alex on the Chapter 1 thread in the Google Group. "Todo" syntax is updated. I also caught a few instances of 
<md>
  mathfoo \text{.}
</md>
that were in the WeBWorK exercises and updated to use
<md>
  mathfoo
</md>.

I think this is ready to merge, and then I'll have a go at chapter 2.

davidfarmer:
chapter_limits.ptx
Line 35  Let's discus if <m>\text{distance} = \text{rate}\times\text{time}</m> is the correct markup
Line 42  period should be outside "term"

sec_limit_intro.ptx 
Line 1347   "1" is an "answer" not a "solution"

sean-fitzpatrick:
There is some work to be done across the whole project with respect to the solution vs. answer question.
Let's discuss how to proceed tomorrow.
In particular: for a WeBWorK exercise, the <solution> division is written to the .pg file. So is <hint>. But not <answer>. I'll have to check the schema to see if <answer> is inside <webwork> or between <webwork> and <exercise>.

sean-fitzpatrick:
Thinking about minimizing effort for merging:
Since changes done prior to this PR are included in the chapter 2 branch, it might be easier to close this without merging. The changes will be captured in the other PR.

sean-fitzpatrick:
Merging #5. Corrections above can be handled in #6

PR #12, opened 2019-06-23
Update Chapter 3

sean-fitzpatrick:
Code from PCC for images and exercises merged into David's conversion of APEX V5.
Also some math formatting, such as using `\fp` instead of `f'<nbsp />`.

It looks like the exercises for curve sketching still need some "WeBWorKing".
Related questions on this:
- Do we want to set up the curve sketch problems so that students have to enter all the relevant pieces, like intervals, critical points, etc?
- Should we give students a Sage cell they can use to check their sketches?

In the PCC version Greg's TikZ images for sign charts were replaced by tables.
I'm not convinced this is an improvement (visually, at least) but I suppose there may have been good reason for it.

One of the changes in my fork of Greg's book is to replace the all-in-one sign chart with separate ones for f, f', and f''. (I always have my students do them one at a time, and add to the sketch with each one.)
Is this something others would consider, or am I the only one who splits them up?


davidfarmer:
Some things don't look right, such as figure inside sidebyside inside figure,
but I am going to merge because this phase is about reconciling previous versions
and later we can make any remaining structural corrections.

PR #13, opened 2019-07-02
Fixed schema violations for chapter 1

sean-fitzpatrick:
I've run validation on the PTX code for chapter 1 to bring things up to date with the latest version of PreTeXt. This includes things like getting rid of uncaptioned figures, changing table captions to titles, etc.

Work involved a combination of manual fixing, plus running the fix-deprecations.xsl utility and then comparing. This did a bit of linting to remove unnecessary spaces in tags, and escaping of special characters. I left characters like &gt; unescaped in tikz and pg-code.

The fix-deprecations utility also removed the <![CDATA[ ... ]]> wrapper from tikz code. Maybe it's not needed any more but I kept it to be safe.


sean-fitzpatrick:
Lyle, I've asked for your review just so you can compare the changes here to what you're doing with the schema violations in the later chapters.

davidfarmer:
I am okay with this pull request, and will merge it later in case someone else wants to take a look.

Several comments:

I don't think an empty "description" tag should be replaced
by a self-closing tag.  We need to put a description at
some point, so this just adds work.  (If a linter did that,
then it would be good to tell it to ignore empty description tags.)

The self-closing "cell" tag also looks odd to me, because it
really is an empty cell and the human eye expects to see
a certain structure.

I suggest that changes like removing spaces before the end of a
self-closing tag should either be their own pull request,
or should be just merged by the person who made that change.
Such a large number of similar changes makes it very difficult
to see the important changes.

If the "greater than" sign is allowed in math mode, then we
should use it (and not & g t ;).  If it is not allowed, then
\gt would be better than the xml version.  Either way, it is
good to have a space on each side of it.
See line 765 of sec_limit_continuity.ptx

The use of "table" is probably incorrect.  For example,
line 226 of sec_limit_intro.ptx .
A "table" in the sense of the Chicago Manual of Style,
has a particular meaning.  A grid of numbers in PreTeXt
is usually *not* a table in the sense of CMOS.

I assume it was intentional that sec_limits_infty.ptx was
deleted.

Some of the above comments apply generally so I will email the group.

sean-fitzpatrick:
Thanks. The self-closing tags for description and cell were changes made by fix-deprecations.xsl.
That's where most of the replacements of > by &gt; come from as well.

The utility is good for things like changing figures to sidebyside, or captions to title, so I'm using it but also trying to work out when I should ignore the changes it suggests.

ptx/sec_limit_intro.ptx
```diff
@@ -307,7 +306,7 @@
             <!-- START figures/fig_limit1.tex -->
             <image xml:id="img_limit1">
               <latex-image>
-              <![CDATA[
+              <![[CDATA[
```
Alex-Jordan: Is it right to insert the extra bracket here? I suspect not. Or, if that is correct, then this block needs an additional close bracket too.

davidfarmer:
I should have mentioned:

It is the job of the person making the pull request to have tested
that the new version works as expected.

Some of my projects have a policy that the pull request has to
specifically state that the code was checked.

I am not suggesting we add that requirement, but I would like that
running the code is the job of the person who makes the change.
Then the person evaluating it only has to look at the source.
(Unless the change in the output is important, in which case the
pull request has to say where to look.)

ptx/sec_limit_intro.ptx
lpsnelgrove: No it causes a crash.

sean-fitzpatrick: That was a last-minute typo on my part. I thought I re-ran validation but missed it somehow.
I'll be more careful on the next one!

sean-fitzpatrick:
Took another pass through chapter 1.
Removed all the CDATA since it doesn't seem necessary now.
Aside from a problem with images that is caused by running mbx on Windows (and not the ptx code) everything seems to be working as expected.

PR #17, opened 2019-07-17
Chapter 4 merge

sean-fitzpatrick:
Update to Chapter 4.
For this chapter, Lyle did some pre-processing of the apex-mbx code to fix deprecations.
Then I merged with the APEXCalculusPTX, making sure to track changes between V3 and V4.
HTML output is at [http://www.cs.uleth.ca/~fitzpat/apex-test/](http://www.cs.uleth.ca/~fitzpat/apex-test/).

Note: text in TikZ images doesn't look right. Could be because I'm using Inkscape instead of pdf2svg for file conversion. I'll try outputting images to PDF to make sure.


sean-fitzpatrick:
I think this material is ready to merge, with the usual understanding that an additional pass through will be needed later on.
For the most part this is a straight merge of APEX V5 with the apex-mbx code (which has been cleaned up by Lyle).

I've included the section on Taylor polynomials here because we teach it at this point in the course.
There's the original version, as well as a modified "early Taylor" version that fits better at this point in the book.

davidfarmer:
Looks good.  Maybe some small changes which can be addressed later.

sean-fitzpatrick:
Thanks. I know there were some images in need of fixing. I've got that coming from Lyle.
Chapter 5 pull request coming right away!

davidfarmer:
I don't rebuild the book (because the assumption is that the person
submitting the pull request has checked that everything works), I just
look at the actual file changes.

So I would not notice a problem with an image.

PR #27, opened 2019-08-04
edited sec_limits_onesided.ptx to remove reference to figure

APEXCalculus:
Edited the one-sided limits section, second example, to remove references to the figure when evaluating the limits. The figure is shown at the end of the example as a reference.

### DIFF
--- a/ptx/sec_limit_onesided.ptx
+++ b/ptx/sec_limit_onesided.ptx
@@ -363,7 +363,8 @@
                        2-x \amp 0\lt x\lt 1 \\
                        (x-2)^2 \amp 1\lt x\lt 2
                     \end{cases}</m>
-                    as shown in <xref ref="fig_onesidedb">Figure</xref>.
+                   .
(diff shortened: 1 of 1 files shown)


sean-fitzpatrick:
This looks good. Having this example will definitely help out my students.
A couple of minor comments from me:
- PreTeXt advice for authors includes trying to minimize use of \displaystyle. Instead of `<m>\ds \lim_{x\to 1^-}f(x)=1</m>` I think the recommended markup would be `<m>\lim\limits_{x\to 1^-}f(x)=1</m>`.
- Some lines are a bit long. I think David's pretty-printing guidelines suggest a maximum of 60 characters per line. (This is minor and can easily be corrected later.)
- I like the verbal explanation. But would a step-by-step calculation also be useful? I might do something like this:
```
<md>
  <mrow>\lim_{x\to 1^-}f(x) \amp = \lim_{x\to 1^-}(2-x) \quad (f(x)=(2-x) \text{ for } 0\lt x\lt 1)</mrow>
  <mrow> \amp = 2-1 = 1 </mrow>
</md>
```

Issue #30, opened 2019-08-07
Little things, mostly punctuation


davidfarmer:
This issue is to collect little things that are easy to fix.

3rd formula after Figure 4.5.1, sentence starting "Let's start with...":  no period at end of sentence.

Formula immediately before Figure 4.5.4:  The "(x)" is on the wrong side of the equals.

Solution to Example 4.5.13.:  no period after first display.  Also at the end of sentence
starting  "Since we are forming..."

Solution to Example 4.5.18.:  Two missing periods at end of sentences (that end in a display).

sean-fitzpatrick:
Fixed. This section usually comes later (Chapter 8). I pulled it in since this is where I'll put it for the Lethbridge version of the book. So I don't think it had seen much hand editing yet. Thanks for the catches.

sean-fitzpatrick:
See the list of 'important points' immediately after Theorem 5.1.6.
It's marked up as a `<ul>` with each `<li>` containing a `<p>` containing an `<me>`.
So the math is centred, with a big gap between the bullet point and the equation.

Probably we should use `<m>\displaystyle ...</m>` here?
It's in use elsewhere. If we're consistent it'll be easier to update if there's progress on the issue regarding lists of equations that came up previously.

sean-fitzpatrick:
In that same list: the discussion on needing `<m>n\neq -1</m>` in the power rule has an unfortunate line break in the equation (on my screen, at least). Maybe put this in display math?

sean-fitzpatrick:
Question 6 in 5.1 doesn't work. I think this is one of several that tries (unsuccessfully) to use <fillin/> in a WeBWorK problem.

sean-fitzpatrick:
I suspect @davidfarmer will have something to say about the comma in Section 5.2 at the end of the 'text equation' `distance = rate_1\times time_1 + rate_2\times time_2`.

(It's done in math mode with \text. There's another one I tried marking up using `<times/>` and regular text. But here the subscripts stumped me.)

sean-fitzpatrick:
Some graphs look wrong in Chapter 5 and later, due to an oversight on my part.
We added some additional global styling definitions in the index.ptx file, and used them to update graphs. This was done on apex-mbx and then merged in, but I forgot to do the index file.

It'll get fixed in the next pull request.

sean-fitzpatrick:
@davidfarmer I'm pretty sure everything here was dealt with long ago. Closing.

PR #32, opened 2019-08-08
Housekeeping

sean-fitzpatrick:
Everything in the housekeeping issue is dealt with except for <ndash/> and \fp vs f'.
Requesting merge so I can make a branch for Chapter 6 with these changes.

Visual inspection of HTML done up to end of Chapter 4. Chapter 5 to come.

Work still needed on WeBWorK problems for optimization and related rates, but that will be a separate issue.


davidfarmer:
Not possible to tell if one of those items was missed, but what I saw looked good.
Just a couple things:

 "differentiable on" not term-ed on line 1264 of sec_deriv_intro.ptx 
 Also consider wording as "differentiable on the interval"

No terms for infinite limits in sec_limit_infty.ptx , but no point in adding those
when we don't have a good way to do it.  Maybe sometimes it is okay to
have math in the term?

sean-fitzpatrick:
Let's add these to the list of "little things" in #30. I'll get them on the next pass.
I'll be opening a branch for fixes to Chapter 5 soon since I've spotted some issues there.

Biggest outstanding work item affecting quality in Chapters 4 and 5 is incomplete WeBWorK code.

PR #33, opened 2019-08-13
Chapter 6

sean-fitzpatrick:
First pass at Chapter 6. WeBWorK code still needs attention, but the same is true in Chapters 4 and 5.

### DIFF
--- a/ptx/chapter_anti_tech.ptx
+++ b/ptx/chapter_anti_tech.ptx
@@ -26,4 +26,3 @@
   <xi:include  href="sec_lhopitals_rule.ptx" />
   <xi:include  href="sec_improper_integration.ptx" />
 </chapter>
-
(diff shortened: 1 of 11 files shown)


davidfarmer:
The formatting changes (prettyprinting) made it difficult to carefully look at the details.

If there is wisdom for when to use `displaystyle lim_{var, thelim}` and when to use
`lim limits_{var, thelim}`, then this should be documented somewhere.

sean-fitzpatrick:
Yes, unfortunately most of these sections had an introduction inserted during conversion that had to be removed. That changes the indentation and ruins all the diffs.
I tried to remember to make that change as a separate commit so looking at the individual commits might make it easier to see details (but then you have to look at all the commits).

As for limits: I think either option produces the same rendering on the limit symbol. The difference is that the former will also apply displaystyle to the function in the limit. So it matters if you have a \frac in the limit, for example. But which to use in that case? I don't have a rule for it.

PR #37, opened 2019-08-30
Conversion of tikz images in sec_def_int to webwork images (and a couple other fixes)

lpsnelgrove:
- Converted every tikz image in a webwork  problem to a webwork image in sec_def_int.
- I converted one of the exercises into a webwork exercise as it had a todo for it and an image to convert but it should be double checked. 
- I also changed over one of the images to use fill between in sec_FTC (others will also be converted over).

### DIFF
--- a/ptx/chapter_integration.ptx
+++ b/ptx/chapter_integration.ptx
@@ -16,13 +16,13 @@
     </p>
   </introduction>
 
-  <xi:include  href="sec_antider.ptx"/>
+  <!-- <xi:include  href="sec_antider.ptx"/> -->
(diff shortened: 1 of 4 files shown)


sean-fitzpatrick:
Looks good. Just one thing to change (that I should have specified -- sorry!)
Rather than deleting the tikz code, better that we comment it out.
That way we can revert to it when/if tikz images are supported in WeBWorK.

You should be able to comment out the whole thing, but watch out for `--` since that will mess with commenting.

sean-fitzpatrick:
Thanks Lyle. I made an adjustment on the image width but this is otherwise great.

PR #40, opened 2019-11-16
Added differential equations content

sean-fitzpatrick:
@davidfarmer did a machine conversion of the differential equations appendix for APEX for me earlier this week, and here is the result after some hand editing.

Adding new content, so no changes other than adding a line to the index file to include the content.

To do/deal with: obviously there is no WeBWorK code at this point. Figures are done, text is done, and it looked pretty good to me when I built the HTML. One schema non-compliant item: in the first section there's an exercise (or set of exercises) that asks the user to match differential equations to slope fields.

I tried to follow the way this was originally, done, with the slope fields labelled a,b,c,d and then 4 exercises following, each with one equation to be matched. But labeling was done by auto-captioning of figures, and figures aren't allowed in exercises.

On the forum, Rob suggested putting the captions into the images themselves, and dropping the figures for simply sidebyside+image. If anyone has a better idea for the markup on this, suggestions are welcome.

### DIFF
--- a/ptx/chapter_differential_equations.ptx
+++ b/ptx/chapter_differential_equations.ptx
@@ -1,0 +1,26 @@
+<chapter xmlns:xi="http://www.w3.org/2001/XInclude"  xml:id="chapter_differential_equations">
+  <title>Differential Equations</title>
+  <introduction>
+    <p>
+      One of the strengths of calculus is its ability to describe real-world phenomena.
+      We have seen hints of this in our discussion of the applications of derivatives and integrals in the previous chapters.
+      The process of formulating an equation or multiple equations to describe a physical phenomenon is called
+      <em>mathematical modeling.</em> As a simple example,
+      populations of bacteria are often described as
+      <q>growing exponentially.</q>
+      Looking in a biology text, we might see <m>P(t) = P_0e^{kt}</m>,
+      where <m>P(t)</m> is the bacteria population at time <m>t</m>,
+      <m>P_0</m> is the initial population at time <m>t=0</m>,
+      and the constant <m>k</m> describes how quickly the population grows.
+      This equation for exponential growth arises from the assumption that the population of bacteria grows at a rate proportional to its size.
+      Recalling that the derivative gives the rate of change of a function,
+      we can describe the growth assumption precisely using the equation <m>P' = kP</m>.
+      This equation is called a <em>differential equation</em>,
+      and these equations are the subject of the current chapter.
+    </p>
+  </introduction>
+  <xi:include  href="sec_Graphical_Numerical.ptx" />
+  <xi:include  href="sec_Separable.ptx" />
+  <xi:include  href="sec_Linear.ptx" />
+  <xi:include  href="sec_Modeling.ptx" />
+</chapter>
--- a/ptx/index.ptx
+++ b/ptx/index.ptx
@@ -291,6 +291,8 @@
 
     <xi:include href="./chapter_app_of_int.ptx" />
 
+    <xi:include href="./chapter_differential_equations.ptx"/>
+
     <!-- <xi:include href="./chapter_sequences_series.ptx" /> -->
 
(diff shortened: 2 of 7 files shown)


sean-fitzpatrick:
Apparently this is where I caught and fixed a typo in an earlier section as well. Sorry for not putting that on another branch.

sean-fitzpatrick:
@Alex-Jordan if you have time/interest at some point, the exercises could use your WeBWorK touch.
@APEXCalculus I think this takes care of the appendix on differential equations. I put it in as a chapter rather than an appendix for now, but it can be moved later.

APEXCalculus:
I think it's fine as a chapter - makes more sense if the expectation is
that people won't use both the print & online versions. Having multiple
print chapter numberings didn't make sense so I added it as an appendix.

sean-fitzpatrick:
Ok, thanks. I put it right after applications of integration. Seemed like the best place for it. I did a build yesterday and it looks decent.

sean-fitzpatrick:
If I can get away from my kids long enough tomorrow I think I can get a build done including this and sequences and series.

Alex-Jordan:
> @Alex-Jordan if you have time/interest at some point, the exercises could use your WeBWorK touch.

I'd love to look through it, and more generally, participate more in this conversion. I'm wrapping up a project obligation this term, and will be "free" for extracurricular activities starting mid January, and realistically that is when I can start being active here.

> makes more sense if the expectation is that people won't use both the print & online versions

At PCC, students and faculty are using both the official APEX print edition _and_ an HTML version that we host (different from the one Sean is updating). It's worth noting that there isn't a separation. On the one hand, you do have people (students and faculty) who are using both versions at the same time. On the other hand, they aren't really upset about the discrepancies. It helps that the faculty who do this are fine with having to deal with homework submissions that may not be what they were expecting.

My point is it would be nice to understand better about what the impact is for HTML and PDF to diverge. I'm not clear if it is a bad thing, even though common sense might suggest it is a bad thing.

sean-fitzpatrick:
That could be an interesting thing to study. For Calc I this semester I'm using APEX V4 for PDF and the most recent APEX PTX for HTML. I haven't heard any complaints from students. But I'm not assigning work from the book -- just recommended practice. Assigned problems are on WeBWorK.

But I'm nearing the point where we might need to start talking about versions again. Having just added sequences and series, I'm again dealing with the fact that our curriculum asks for Taylor Polynomials in Calculus 1 (right after differentials), but you ideally want to reprise that content before doing Taylor series.

In HTML, no problem -- the Taylor series section xrefs back to Taylor polynomials, and everything is there in one place. But in print, you'd like to include the content in both your Calc 1 book and your Calc 2 book, and avoid multiply-defined references, etc.

sean-fitzpatrick:
@Alex-Jordan on the Google group there's a thread I started awhile ago with some to-do items.
If you have time to tackle some of them, that would be great.

There are several WeBWorK-specific items.
A little one is that PreTeXt markup for units doesn't work in WeBWorK exercises. (Also, now that there's updated character encoding, can we finally use \theta as a WW variable?)

A big one is getting the exercise groups to work right for both static and interactive. And implementing subexercises, I guess.

Another big one is getting TikZ code to work in the exercises. But there seems to be progress on that front? [https://github.com/openwebwork/pg/issues/418](https://github.com/openwebwork/pg/issues/418)

Alex-Jordan:
Copy that.

I've been meaning to mention to you that the mbx script now can handle WW 2.15. I forget if you still have some custom workaround on your WW server, reverting that one character to \x85 or something like that. But an up to date PTX should handle things fine now, whether the WW server is using \x85 (what it used to use), \x1E (what it was using in the beta for 2.15), or \x0D (what it uses in the final 2.15, and going forward).

sean-fitzpatrick:
I think we're up to date on our WeBWorK server. In any case we're replacing it once the term ends. New machine is set up and ready to go.

PR #41, opened 2019-12-26
Chapter 9

sean-fitzpatrick:
Chapter 9 merged and cleaned up.
Some tables needed a lot of work. Images mostly good, thanks to earlier work by Lyle.

WeBWorK problems still need some attention, especially with instructions for individual problems.


sean-fitzpatrick:
A lot of problems had webwork ptx tags wrapped around them, but no pg-code.
I commented out the webwork markup since there's no real benefit in sending these to the webwork server to get the same text back. Need to do this for Chapter 8 as well.

PR #48, opened 2020-03-26
WeBWorK pass-through

Alex-Jordan:
I need to use the APEX WW exercises in my class (which will cover chapters 1--3, and section 4.2). As I go, I intend to clean up the source, possibly restructuring a problem so it looks better in all output modes. Not changing what the problem asks for. 

I did add two problems to section 1.1 here, hoping that is OK. I will refrain from doing that now that I realize it would affect numbering for you (Sean) if you are using these right now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,7 +123,7 @@ pdf:
 	cp -a $(WWOUT)/*.png $(PRINTOUT)/images || :
 	cp -a $(IMGSRC) $(PRINTOUT) || :
 	cd $(PRINTOUT); \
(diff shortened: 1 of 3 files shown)


sean-fitzpatrick:
Hi Alex,

This is welcome and arguably overdue attention -- I haven't had the capacity to do it myself.
Don't worry about changing exercise numbering. I don't assign textbook problems for homework.

sean-fitzpatrick:
Not merging yet in case you have further changes to push. If not, go ahead and merge

Alex-Jordan:
I have a more systematic way coming up to deliver all this.

PR #49, opened 2020-05-17
Recent work

Alex-Jordan:
I'll merge this in a day or so, unless @sean-fitzpatrick does first. I thought I should leave an opening for looking things over.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,7 +123,7 @@ pdf:
 	cp -a $(WWOUT)/*.png $(PRINTOUT)/images || :
 	cp -a $(IMGSRC) $(PRINTOUT) || :
 	cd $(PRINTOUT); \
-	xsltproc -xinclude $(PTXXSL)/mathbook-latex.xsl $(OUTPUT)/merge.xml > apex.tex; \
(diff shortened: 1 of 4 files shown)


sean-fitzpatrick:
Looks good. Have you done much work on other sections?
I think there's a lot of cleanup (things like getting rid of `seed="1"`) that I did on the videos branch and forgot to merge.
If you don't have too much uncommitted work in progress I can prioritize merging any changes to the first couple of chapters to reduce chances of conflict later on.

Maybe @APEXCalculus has an opinion on changing `\lim\limits_{}` to `\lim_{}`. Greg made much more liberal use of displaystyle in the original LaTeX.

Alex-Jordan:
I've done a lot more with chapters 1 and 2, but now I will be going through the checklist spreadsheet I sent you and submitting one section at a time.

Things like the `lim` observation are what I'm looking for from a quick look-over by someone else's eyes. 

If we are trying to make a certain level of print quality, it should not be `\lim\limits` in regular sentences. That causes a change in the distance between one line's baseline and the next. There are consequences when you have a printed two-page book. One is that the left page and the right page have lines of text that are not co-aligned. The typesetter's guide I use recommends avoiding this. I know that many people will shrug.

The other is that the front and back of a page have lines of text that are not co-aligned. The guide also recommends avoiding this, but I've seen firsthand how this one matters for an OER. When the paper and ink quality are lesser (which may be the case with a student printing their own) then you can see through to the back side to a degree. When lines are not co-aligned, this results in distractions present in the white space between two lines.

So you have the arguments for not using `\limits` in the middle of a sentence. Take them under consideration, and if it's still important to use `\limits`, I'll change these back.

Alex-Jordan:
I'm going to merge this, with the awareness that @APEXCalculus may want the `\limits` to come back. Not hard to do.

PR #50, opened 2020-05-18
Revisions to Section 1.2

Alex-Jordan:


sean-fitzpatrick:
Everything looks good, although I didn't check the exercises carefully - the diff blows up a bit there. Wait for @APEXCalculus  to approve the changes to the list before merging?

Alex-Jordan:
@APEXCalculus I wanted to double check you saw the discussion about the 3' and 3'' list. The change here goes a step beyond what I'm comfortable just doing without your approval.

Some not entirely up to date output is here:
http://spot.pcc.edu/~ajordan/temp/apex.pdf#page=10
http://spot.pcc.edu/~ajordan/temp/sec_limit_def.html

sean-fitzpatrick:
One thing I'm just noticing now: the TikZ image containing line 272 is much improved.
But I noticed you're using `|->` instead of `|-&gt;`.
Do we want to escape the `>` just to be safe?

Alex-Jordan:
I don't escape `>` because it's tedious, especially in WeBWorK where the perl has lots of `->` and `=>`. (It's much more rare in perl to need `<`.)

We're fairly sure that it is unnecessary to escape it. But I don't object of you'd like to escape those. 

Note sometimes I'll escape `--` inside tikz code, because `--` is not legal within an XML comment, and so if you ever need to comment out a large block containing that "image", having the raw `--` is a problem.

PR #51, opened 2020-05-19
Change list format; escape some less than signs

Alex-Jordan:


sean-fitzpatrick:
Check with @APEXCalculus on the change to the limit property for compositions. I think the change here reverts to APEX V4 -- but I can't recall now if the added details were yours, or Greg's. (I checked the Google groups and found a discussion on continuity of composition -- which we never really resolved! -- but not limits.)

Alex-Jordan:
@APEXCalculus should check on this, but I reviewed that same thread and I think what happened was:

1. Version 3 had an incorrect version of the theorem.
2. The PCC conversion attempted to fix it by making the version with two options.
3. Greg mulled it over and just took the first option when making the non-PTX edition 4.

So here I thought to go with that one.

APEXCalculus:
I must be doing something wrong - when I looked at Alex-Jordan:master PTX files, it seems the composition of limits portion of the theorem is from V3, which was wrong. So I can't tell what the new limit statement is. Is it just what appeared in V4? 

This is what I see in Alex's ptx folder:
<p>
              Adjust the limit requirements to
              <md>
                <mrow>\lim_{x\to c}f(x)\amp=L\amp\lim_{x\to L}g(x)\amp=K\amp g(L)\amp=K</mrow>
              </md>.
              Then <m>\lim\limits_{x\to c}g(f(x)) = K</m>.
            </p>

sean-fitzpatrick:
I think this is fine. The part that was missing in V3 is g(L)=K (continuity of g at K). The previous ptx version had some additional condition that's sufficient if g is not continuous, but likely to cause confusion for students.

Alex-Jordan:
@APEXCalculus It sounds like you are looking at the right thing/place, but did not see that what's there has the post-version 3 correction. It's the same as version 4, except the phrasing "Adjust our previously given limit situation to" became "Adjust the limit requirements to". The actual conditions are what they should be to make a useful and correct theorem.

sean-fitzpatrick:
One note/question in `sec_limit_analytically.ptx` as I compare to the videos branch: in `<theorem xml:id="thm_lim_continuous">` there is the change (for example) from
```
<li><m>\lim\limits_{x\to c}\cos(x) = \cos(c)</m></li>
```
to
```
<li>
  <p>
    <m>\lim\limits_{x\to c}\cos(x) = \cos(c)</m>
  </p>
</li>
```
Would it make more sense to just drop the `\limits` (it's redundant)?
Note that for an `<m>` inside an `<li>` that is not wrapped in a `<p>` we now get automatic display style.

So if the intent is to remove displaystyle here, we should also remove `\limits`.
But I think the list looks better with displaystyle.

(Whether or not we keep teh `<p>`, we probably should remove `\limits`.)

Alex-Jordan:
Last year in making ORCCA's PDF, I saw spacing differences between list items that use `p` and those that do not. Maybe that is a bug and should be looked into. But I'm in the habit of making all the `li` into the same format within the same list for that reason.

I lean towards keeping the `\limits` in this situation where it's somehow really a part of what the author is doing, and not just styling. But I don't care enough to change these one way or the other. The automatic displaystyle is great.

PR #54, opened 2020-05-23
WW revisions to 1.4

Alex-Jordan:


Alex-Jordan:
This now goes up through all of chapter 1.

HTML: https://spot.pcc.edu/~ajordan/temp/APEXchapter1/chapter_limits.html
PDF: https://spot.pcc.edu/~ajordan/temp/APEXchapter1/apex.pdf
PG: https://webwork.pcc.edu/webwork2/apex-demonstration

`git log master..origin/calc1-videos ptx/sec_limit_* > temp.txt` reveals there are 18 commits in `calc1-videos` that affect sections in chapter 1, and the commit is not in `master`. I could start manually looking at each one and applying what matters to new commits in `master`. I suspect it can't be automatically merged because much of what I have done recently has affected line breaking and indentation in many lines.

sean-fitzpatrick:
I will have time to work on this today. (Probably afternoon -- my wife prefers the morning for working and someone has to watch the kids. I get 5 minutes to work now before they wake up... and as I type this I hear footsteps upstairs... better hurry.)

What I have been doing (which is probably not best git practice, but I think the outcome is the same) is to use Meld to compare the two branches, and just copy over changes as needed. (So, one file at a time rather than one commit at a time.) I can change this workflow if there are reasons not to do it this way.

One comment: at 09a2852eb7933051a896fa0861a86bd8a979692d I notice that you removed `valign="bottom"` from the first `sidebyside` (and others).

That change jumps out visually for me in the HTML: compare [your version](https://spot.pcc.edu/~ajordan/temp/APEXchapter1/sec_limit_intro.html#p-9) to [my version](http://www.cs.uleth.ca/~fitzpat/apex-calculus/sec_limit_intro.html#p-9).

To me, it looks better with the captions aligned. But maybe this causes problems in PDF? I assume there was a reason to remove the vertical alignment.

Kids are up now. I'll look at the other changes more carefully later.

sean-fitzpatrick:
Actually elsewhere in that commit you added `margins="0"` but kept the alignment attribute. So maybe this was not intentional.

Alex-Jordan:
The images should be the exact same dimensions, and so `valign="bottom"` should be better. If you are viewing one of the pages where the valign is bottom, then try zooming in a little with the browser's regular zoom. You will reach a point where the second caption wraps to a second line, and then the caption alignment is gone. If these are top-aligned (and the images match in size) that is not a concern.

But what I cannot figure out is why the first image is smaller than it should be. It is getting left and right margins that take up space within the allotted 47%, making the image more narrow. Then in reduces vertically proportionately, and that lack of alignment happens. @davidfarmer, any ideas why those margins are there around Figure 1.1.1 [here](https://spot.pcc.edu/~ajordan/temp/APEXchapter1/sec_limit_intro.html#p-9)?

Alex-Jordan:
Oops, I meant `valign="top"` (the default) should be better.

Alex-Jordan:
OK, I see a clue. The first caption is very short. If I make that caption longer then the image width grows too (the margin around the figure is reduced to zero). Something about the `margin:auto` is not calculating the margin like we would like it to.

Even when the images reach the same width, then they are not the same height. I thing that is because in the second image, the tick labels extend beyond the default width of a pgfplots image, growing the original dimensions unevenly. `pgfplots` can be set to still reduce everything so that the original default width (which I think is 240 pt) is preserved. I'll look into that. Maybe the margins issue will go away because this other thing might affect the browser's auto margin calculating algorithm.

sean-fitzpatrick:
I think not all images have exactly the same aspect ratio. So setting dimensions by width can cause some mis-alignment.
If I'm the only one who doesn't like the look of off-set captions, I can live with it.

What about the sidebysides that have an image next to a (relatively small) tabular? For example, [the one in this example](http://www.cs.uleth.ca/~fitzpat/apex-calculus/sec_limit_intro.html#ex_limit2)

I didn't like the way these looked without adding `valgin` to the bottom.

sean-fitzpatrick:
But as far as this particular commit: I don't see any concerns here. I think it's safe to merge.

Alex-Jordan:
What I _think_ happened (but have not yet verified) is that a pgfplots `axis` graph is 240pt wide by default, and yes, it has a fixed aspect ratio. These dimensions are used to fit the xmin..xmax and ymin..ymax viewing window. If you notice though, tick labels extend past the xmin..xmax and ymin..ymax viewing window. I think that pgfplots is leaving the viewing window 240pt wide and allowing the tick labels to extend past that, widening the final cropped image. And the same down under the x-axis, but to a different degree. The result breaks the aspect ratio. This may all be controllable with a global pgfplots setting.

Separate from all that, I can see in the Firefox inspector that the left figure has left and right CSS margins while the right one does not. It may be true that these will go away once the aspect ratio on the images is corrected. However, I still think now that @davidfarmer should examine those margins. Because if we legitimately had two images of differing aspect ratios, we should not have those margin there.

sean-fitzpatrick:
OK. Since that particular figure was merged on the last PR, maybe we make this an issue and come back to it? I have about an hour right now to get some work done.

If I merge this I can dive into a manual comparison between master and the videos branch and have that done. After that I'll consult with one of you with more git-spertise than me on how to merge future changes into the videos branch without overwriting the videos.

Alex-Jordan:
I don't mind taking a deeper look at the chapter 1 bits and seeing if I can find a smooth git way to merge them, or at least assessing how much will go into a manual merging. Let me try that first.

I'm optimistic about just merging chapter 4 and beyond directly. Maybe 2 and 3 as well, since it's possible I only did WW edits there (can't recall right now).

sean-fitzpatrick:
OK. Added the issue. I've gotten pretty fast at manual merging in Meld since that's how I resolved most of the differences between this version and teh apex-mbx version.

I added #55 to pull in all the changes for Chapters 4-10. I can do 11-15 first while you see if you can find a more elegant way of reconciling Chapters 1-3

sean-fitzpatrick:
Question about the `pg` graphs in WW exercises:

I noticed that you changed the size defined in the `pg-code`, and also that you removed the width specification on the sidebyside. Is there a default width being applied now?
The static webwork problems don't look good if the image width is too big.

Alex-Jordan:
If they are in an exercisegroup with cols="2", then we want them at 100%. Holler if you see an example where it still looks bad in output. Have you tried using the Makefile to generate output now? After setting up `Makefile.paths`, then run
```
make webwork-extraction
make merge
make pdf
make html
make pg
```

To start out, comment out all but chapter 1 just to see if it's working.

sean-fitzpatrick:
Ahh, of course! I didn't catch the 2 column exercise groups.
Do the two columns become one on mobile?

I haven't tried the makefile yet.
Right now I'm taking a look at Chapter 1 to bring the videos branch even with your changes.
So far I haven't found much on my branch that needs to be committed to `master`. I think for the first few chapters I had time to merge changes (probably at the end of the fall semester).

PR #59, opened 2020-05-28
clean up some subsection xml:ids

Alex-Jordan:
I'm starting to bring in the things from chapters 1--3 in `calc1-videos`. They will accumulate here. No need to merge until I'm done.

### DIFF
--- a/ptx/sec_Modeling.ptx
+++ b/ptx/sec_Modeling.ptx
@@ -12,7 +12,7 @@
       <idx><h>differential equation</h><h>modeling</h></idx>
     </p>
   </introduction>
-  <subsection xml:id="fig_20_translation">
+  <subsection>
     <title>Models Involving Proportional Change</title>
     <p>
       Some of the simplest differential equation models involve one quantity that changes at a rate proportional to another quantity.
@@ -21,7 +21,7 @@
       The words in this assumption can be directly translated into a differential equation as shown below.
     </p>
 
-    <figure>
+    <figure xml:id="fig_20_translation">
       <caption>Translating words into a differential equation</caption>
       <image xml:id="img_20_04_translation" width="40%">
         <description></description>
--- a/ptx/sec_limit_infty.ptx
+++ b/ptx/sec_limit_infty.ptx
@@ -495,7 +495,7 @@
     </p>
   </subsection>
 
-  <subsection xml:id="fig_hzasy">
+  <subsection>
     <title>Limits at Infinity and Horizontal Asymptotes</title>
     <p>
       At the beginning of this section we briefly considered what happens to
--- a/ptx/sec_limit_intro.ptx
+++ b/ptx/sec_limit_intro.ptx
@@ -959,7 +959,7 @@
     </example>
   </subsection>
 
-  <subsection xml:id="fig_diff_quot_small_h">
+  <subsection>
     <title>Limits of Difference Quotients</title>
     <p>
       We have approximated limits of functions as <m>x</m> approached a particular number.
--- a/ptx/sec_riemann.ptx
+++ b/ptx/sec_riemann.ptx
@@ -389,7 +389,7 @@
     </example>
   </introduction>
 
-  <subsection xml:id="fig_rie_notation">
+  <subsection>
     <title>Summation Notation</title>
     <p>
       It is hard to tell at this moment which is a better approximation: 10 or 11?


Alex-Jordan:
In this one commit so far, what happened is I saw a place where you cleaned up an xml:id for a subsection in section 1.1. I globally searched for the same issue and ended up making 4 such corrections. Two of them are outside of chapters 1--3.

Is it OK for me to work that way? Or would you prefer I stay completely within chapters 1--3?

Alex-Jordan:
OK, after reviewing all commits from `calc1-video` that touched chapter 1, the subsection xml:id cleanup was the only thing not video-related. On to chapter 2.

And yeah, now I see lots of paragraphs like "In the following video, you will see...". Those will make it hard to get a clean new video branch. Once we are done with the current phase (getting a clean `master` that includes non-video things from `calc1-videos`) we should try merging `master` into `calc1-videos` and see if we can run with that. It may result in some issues on that branch but they could be patched directly over time. And it would be a lot less effort than other things I can imagine.

sean-fitzpatrick:
I think if you find a fix that should apply globally and it hits later chapters, that's fine.
I'll start working on this tonight. No opening for daytime work today...

I've got a copy of the branch with the merge attempt.

If I compare against master it's easy to see the extra paragraphs that introduce videos. There aren't many of these -- at most 10 I think. I think what I'll do is comment them out, but also add an xml:id in those paragraphs that tags them as related to a video.

sean-fitzpatrick:
How do you want me to proceed with merging commits that affect non-video content on the calc1-videos branch?

How about this: instead of the recently closed pull request into `master`, we:

- make a clean branch off of master
- make the pull request into that branch instead of master
- go through that branch to look for anything that went wrong in the merge
- make changes as needed
- merge that branch back into master

Or does that last step just run into the same trouble again?
Another option is to keep that clean branch as a backup copy, and merge into master.
Then any errors we find can be fixed with commits directly into master.

Alex-Jordan:
> How do you want me to proceed with merging commits that affect non-video content on the calc1-videos branch?

First, one thing I noticed is that just because a commit message is "did such and such regarding videos" that doesn't mean the commit doesn't have non-video-related edits. This is why I'm catching things as I go through the commits one by one in chapters 1--3. Just making sure we are on the same page.

I figured I'd leave it to you to handle for chapters 4+. It sounds like you have a tool that makes it easier. I just want to handle 1--3 since I did recent work there. If you open some PR into master that only affects files from chapter 4 and beyond (and does not add video-related content) then I can review that PR. (I think this is what #55 was doing, and I just hadn't gotten to reviewing it.)

With chapters 1--3, I am not using git to make the edits. I am using git to see what edits would be lost if we cut `calc1-videos` away, then making those same edits manually on `master` and then committing new like the one commit that is here.

Alex-Jordan:
For example, this commit is called "Add videos on derivatives" but it does a lot more to `sec_deriv_chainrule.ptx`:

https://github.com/APEXCalculus/APEXCalculusPTX/commit/5e52a74b9eba5ed32e65839606f0a73c0433e822

Currently trying to convince myself it's just indentation.

Alex-Jordan:
I got through chapter 2, found a few things. I'm going to merge this, and start doing the full exercise review merging for chapter 2.

sean-fitzpatrick:
Ugh. Yes -- I removed an introduction that didn't belong. Which absolutely should have gone in another commit. That said, if I compare what's on the videos branch now to the current master, there are only three non-video differences, other than cleaning up the webwork tags:

On line 13: `<xref ref="thm_PowerRule" text="title">Power Rule</xref>` in `master`, and `<xref ref="thm_PowerRule" text="title/>` in `calc1-videos`.

I think the latter is correct, and was changed after Rob pushed an update to the `xref` code that started giving me warnings.

At line 198 in `master`, the second example reads,
"Use the Chain Rule to find the derivatives of the following functions,
        as given in <xref ref="ex_chain1">Example</xref>."

But then no functions follow. So I rewrote that to say,
"Use the Chain Rule to find the derivatives of the functions <m>F_1(x)</m>,
        <m>F_2(x)</m>, and <m>F_3(x)</m>,
        as given in <xref ref="ex_chain1">Example</xref>."

And later on there is one `displaystyle` for inline math in `calc1-videos`  that's been removed in `master`.

sean-fitzpatrick:
OK -- good news (I think). For chapters 4 and 5, if I compare your `merge-attempt` pull request against my `errata` pull request, (with `calc1-videos` as a third point of reference) the only difference between our pull requests is that you commented out the videos, and I removed them. (Other than a couple of paragraphs that are tied to the videos -- I removed those as well.)

PR #60, opened 2020-05-29
remove legacy files

sean-fitzpatrick:
This gets rid of some legacy files from the original LaTeX files from APEXCalculusV5: WeBWorK files (these are the old .pg files done by Shaun van Ault), and header files (LaTeX formatting).

I was going to also remove `exercises/` and `figures/` but these are still in all branches.

Should make it easier to compare branches if these files are gone from `master`.


sean-fitzpatrick:
Merging. This just removes unnecessary files that we can pull from APEXCalculusV5 in any case.

PR #63, opened 2020-06-01
Bring Chapter 4 level with videos branch

sean-fitzpatrick:
- Three global commits: remove `seed="1"`, remove `<setup>`, change `\ ` to `\,`.
- Videos included but commented out.
- A few video intro paragraphs identified, given `xml:id`, and commented out.
- Whitespace cleaning.
- A handful of small corrections.


sean-fitzpatrick:
Also included fix to back reference.

sean-fitzpatrick:
@Alex-Jordan the commits applied across the project hit Chapter 2 as well. But they match (I think?) changes you also have to remove the seed spec and the setup tag.
Let me know if you want the first three commits split into a separate PR.

Alex-Jordan:
I merged mine just to see, and it made it so this one conflicts. I think it will be best on your end to leave out anything from chapters 1--3.

Alex-Jordan:
Edits I have made in chapters 1--3 include:

- indentation (a mistake, but it's too late now)
- small edits to the content (which is why I can't just ignore my indentation edits)
- substantial edits to WW code (which happen in conflict with the seed/setup edits you had made.)

BTW, all-out removal of the seed specification is not entirely the right move. There are places where it is 100% needed. These are the places where "exercises 20--29 use the same functions as exercises 10--19", etc. 

As I go through, I am leaving `seed="1"` wherever there is any actual randomization and conditioning random values on `seed==1` to keep the exercises matching the edition 4 values. Because why not. It will please some adopters. And the issue that we had with graphs will be resolved once I can tend to it using information from Mike.

sean-fitzpatrick:
Ok. I'm going to checkout the files for Chapters 1-3 into my branch and then merge. Can you check my changes to index.ptx to see if you have any concerns there?

Alex-Jordan:
OK, the index changes look fine.

You may not have success with just checking out those files and merging. Because the git history on this branch will still have to pass through this conflict. It's possible GitHub can "see to the end" and overcome that, but I don't know.

If that turns out to be the situation, then I can suggest a different approach. (I may even just try it using your branches, since it's the kind of thing where "doing" would be easier than writing it down.)

sean-fitzpatrick:
I'm open to suggestions if you have them. My turn to work for the day starts in about 15 minutes. :-)

I could try to undo the first few commits, or I could manually force the differences in the first 3 chapters from master onto my branch as a new commit. (If the files end up the same but have different commit histories, do we avoid the conflict?)

Alex-Jordan:
> If the files end up the same but have different commit histories, do we avoid the conflict?

I don't think so. Because if someone were to then pull the branch, their local git rebuilds it by progressing through commit patches. (But maybe I misunderstand how git works.)

> My turn to work for the day starts in about 15 minutes. :-)

You mean your turn to do APEX work or domestic work? I just got out of a class, gonna make lunch for the family. Actually, minus one, since our 3-yr-old restarted preschool today. Then class at 2--4:20, but then pick up the kiddo, make dinner, etc. All that is to say I won't be doing much APEX today.

sean-fitzpatrick:
I'm starting work now on APEX and the like.
I can try to revert those global commits, or if you have a better idea I can defer to you.

sean-fitzpatrick:
OK, proceeding as follows:

Running `git rebase -i HEAD~17` to drop the three global commits. That gets me merge conflicts in the Chapter 4 files but those are easy to deal with. Should be done in 5 minutes.

sean-fitzpatrick:
Shoot. That failed. Maybe easier to just redo. Not many changes in Ch. 4 to commit anyway.
I'll try again.

PR #65, opened 2020-06-02
Bring in changes to Ch. 5 from videos branch

sean-fitzpatrick:
Main changes are adding videos and removing deprecated code from webwork exercises.
Last two commits have more edits to the sections on substitution and hyperbolic functions.

### DIFF
--- a/ptx/sec_FTC.ptx
+++ b/ptx/sec_FTC.ptx
@@ -1461,12 +1461,10 @@
       </introduction>
 
       <exercise>
-        <webwork seed="1">
-            <setup>
-
+        <webwork>
             <pg-code>
             </pg-code>
-            </setup>
+            
             <statement>
               <p>
                 How are definite and indefinite integrals related?
@@ -1480,12 +1478,10 @@
       </exercise>
 <!-- Exercise 2 -->
       <exercise>
-        <webwork seed="1">
-            <setup>
-
+        <webwork>
             <pg-code>
             </pg-code>
-            </setup>
+            
             <statement>
               <p>
                 What constant of integration is most commonly used when evaluating definite integrals?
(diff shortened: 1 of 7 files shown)


Alex-Jordan:
I did not scour, but I skimmed, and nothing jumped out.

Except maybe using `\implies` instead of the `\Rightarrow` with hard coded spacing. But if that is a good change, it can come later.

sean-fitzpatrick:
Changing to `\implies` sounds good, but yes -- maybe should be done globally later. Here (and throughout, in later chapters, I think) `\Rightarrow ` is common.

A couple is the display equations that use it seemed cluttered to me so I added some manual spacing.

Alex-Jordan:
One thing to note. This branch `ch5-update` was on the `APEXCalculus` GitHub repo, and I think it would make more sense for these branches to be on our personal repos, then merged into `master`. That would keep out "branching clutter" from the `APEXCalculus` repo. Not so much a concern if it's just 2 or 3 of us, but in the future if there are more contributors, it could be confusing.

sean-fitzpatrick:
Fair enough. I haven't usually been forking. Partly in case more than one person is working on a branch, and partly because it took me awhile to figure out how to keep my fork even with upstream...

I've been using the workflow that was preferred on other projects I've worked on: branch, work, merge, delete branch. But I can do that on my own version.

I could keep the videos branch on a personal fork as well, but I know that last year there was some discussion about not wanting a proliferation of forked versions of the book.

Alex-Jordan:
> I could keep the videos branch on a personal fork as well, but I know that last year there was some discussion about not wanting a proliferation of forked versions of the book.

I could be wrong, but I don't think that is literally about git forks. I think that is about conceptual forks, like "Sean's video version of APEX" and "PCC's early trig version of APEX".

PR #68, opened 2020-06-02
Update Chapter 6

sean-fitzpatrick:
Mostly:
- add videos
- remove deprecated webwork stuff

But there are some typo fixes and minor changes to layout and presentation in the last commit.

### DIFF
--- a/ptx/sec_IBP.ptx
+++ b/ptx/sec_IBP.ptx
@@ -779,7 +779,7 @@
       </introduction>
 
       <exercise>
-        <webwork seed="1">
+        <webwork>
             <setup>
 
             <pg-code>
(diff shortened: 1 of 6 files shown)


sean-fitzpatrick:
Merging. I'm trying to make `calc1-videos` even with `master` as I go. So hopefully at the end the main difference is whether or not the videos are commented out.
A couple of details:
- where there is a paragraph introducing the video, I've added an `xml:id` starting with `vidint-` inside the `<p>` tag.
- videos inside solutions are now standalone (no figure)
- videos outside solutions are still in figures. For these I have the `xml:id` on the figure rather than the video. Maybe that should change?
- All `xml:id`s associated to videos begin with `vid-`

PR #72, opened 2020-06-06
Chapter 10 update

sean-fitzpatrick:
And here is Chapter 10, on plane curves.

This is the last of the chapters from my previous merge attempt. Chapters 11-15 still to go.
But those chapters have a number of corrections to make first.
To cut down on number of pull requests, I'll make changes on the videos branch, build, check, and then follow the same process here to bring the branches even (modulo videos).

### DIFF
--- a/apex.xml
+++ b/apex.xml
@@ -1,0 +1,56 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<!--********************************************************************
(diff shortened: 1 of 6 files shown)


sean-fitzpatrick:
The publisher file isn't part of the Chapter 10 update but I figured I'd add it before I forget

PR #79, opened 2020-06-17
Ch14 update

sean-fitzpatrick:
A number of small changes, including some that I missed in Chapter 13:
- remove periods on captions
- replace `<ndash/>` by hyphen
- replace 'i.e.' by `<ie/>`
- fix scientific units
- add in some videos

Still to do (don't merge yet!):
- many figures that are in `<solution>` should be in the statement of examples.

### DIFF
--- a/apex.xml
+++ b/apex.xml
@@ -34,13 +34,13 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
         <!-- @colors now implemented                               -->
         <!-- also @knowls, @toc, and @banner, although no          -->
         <!-- alternative styles exist yet for those.               -->
-        <css style="default" colors="blue_red"/>
+        <!-- <css style="default" colors="blue_red"/> -->
         <!-- Magic IDs identify HTML pages to analytics services   -->
         <!-- Presence implies relevant Javascript will be added    -->
         <!-- StatCounter requires both values to be set            -->
-        <analytics google-gst=""
+        <!-- <analytics google-gst=""
                    statcounter-project=""
-                   statcounter-security=""/>
+                   statcounter-security=""/> -->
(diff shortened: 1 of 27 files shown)


sean-fitzpatrick:
OK, the last commit is a big one (sorry!) but important, I think:
In Chapters 1-5, we're quite consistent about placement of figures in examples: if the statement of the problem refers to a figure, that figure should go in the statement, not in the solution.

I think it's possible that someone else might disagree on this point, so I'll let @Alex-Jordan or @Alex-Jordan comment before merging. But that's certainly how we did things in the first 5 chapters.

In later chapters the figures sometimes end up in the solution. I think this more or less comes down to where in the original LaTeX source Greg had the `\mfigure` code that placed the figure. In later chapters, sometimes the figure is in the statement; sometimes it's in the solution. I thought we should at least be consistent.

So if a figure gets referenced in the statement, I put it there. I think this improves the visual appeal of the book: the figures look good, so don't hide them behind a knowl :-)

I found misplaced figures going back to Chapter 7 or 8 I think, so this pull request now hits more than just Chapter 14. (But it still doesn't touch the earlier chapters Alex was working on.)

sean-fitzpatrick:
(Didn't mean to tag you twice Alex! Fixed the comment to tag Greg as well.)

Once this PR goes through, I'll tackle the last chapter. I think the two branches should be close to even there. But I need to go through to see if there are any corrections to be made. (And I might have some videos to add.)

sean-fitzpatrick:
I think this can be merged now. I'll leave it for a few more days. Further changes are coming down the pipeline but they can sit in the videos branch for a bit.

sean-fitzpatrick:
Merged. Last chapter coming as soon as I have a couple of hours to spare.

PR #86, opened 2020-09-04
WW revisions and updated Makefile

Alex-Jordan:


Alex-Jordan:
Planning to complete WW revisions to chapter 2 and 3 here. This was all started back in spring. Then I can break and try my hand at organizing a "main" branch for the APEXCalculus repo to be the official main branch. It will involve elements of this repo's master, and your (@sean-fitzpatrick) branches.

sean-fitzpatrick:
Sounds good. Not counting these changes (and possible oversights by me) I think everything that is in `APEXCalculus:master` is also in `sean-fitzpatrick:main`. But that branch has a number of commits not in `master`.

More significant differences:
- videos: commented out in `master`. But if we merge the stylesheets from my `main`, this isn't necessary.
- leftover files: I got rid of a lot of the remaining files from the LaTeX version that weren't needed anymore.

I just did a cleanup of the Asymptote code to get rid of the "sample code". But that doesn't affect Chapters 2 or 3.

sean-fitzpatrick:
Questions about the publisher file. (I noticed you deleted mine and added your own.)

- Should the Runestone option be commented out? I don't think we're using Runestone as the default build, are we?
- If we are going to load Google Analytics, should we first set up a warning about cookies? (This certainly leads to tracking cookies, no?)
- Is that your analytics account? I had one but stopped using it years ago. We probably want whoever is hosting the book to have access to the analytics.

Alex-Jordan:
Sorry, the publisher file was copied form ORCCA and I didn't proofread carefully. I'll adjust. Please don't merge this until I do.

Alex-Jordan:
OK, force pushed.

sean-fitzpatrick:
Is it ready for merge? I'm busy until later tonight (like 10pm) so you have time to make additional changes.

I'm going to put up a draft build of the PDF later to see what people think. So far, so good.
One thing: we need a mechanism for exempting a figure from being placed in the margin.
So far this only matters for the number line figures in Chapter 3. They're meant to be placed inline and don't fit in the margin.

We could add an attribute, or maybe just change the markup a bit? (E.g. if we put those figures in a sidebyside I think that exempts them.)

ptx/sec_deriv_prodquot.ptx
```diff
@@ -440,25 +441,16 @@
...
-        <ol cols="2">
-          <li>
-            <p>
-              <m>\displaystyle \lzoo{x}{\sin(x)} = \cos(x)</m>
 
-      <idx><h>derivative</h><h>trigonometric functions</h></idx>
-
-            </p>
-          </li>
+          <idx><h>derivative</h><h>trigonometric functions</h></idx>
```
sean-fitzpatrick, line 445: Can `<idx>` go here? I thought it had to be inside a `<p>`.

ptx/sec_deriv_prodquot.ptx
```diff
@@ -440,25 +441,16 @@
...
 
+      <p>
+        <ol cols="2">
+          <li><m>\displaystyle \lzoo{x}{\sin(x)} = \cos(x)</m></li>
```
sean-fitzpatrick, line 449: The `\displaystyle` is now redundant. I must have missed this one.

ptx/sec_deriv_prodquot.ptx
```diff
@@ -518,10 +510,12 @@
     <statement>
       <p>
         Find the derivatives of the following functions.
-
         <ol>
-          <li><m>f(x) = \dfrac{1}{x}</m></li>
-
+          <li>
```
sean-fitzpatrick, line 514: Maybe `<li><m>f(x)=\frac{1}{x}</m></li>` is best? Without `<p>` it will insert displaystyle automatically.

ptx/sec_deriv_prodquot.ptx
```diff
@@ -711,1446 +697,1395 @@
     such as <q>the derivative of <m>\sin(x)</m> is <m>\cos(x)</m>.</q>
     The <xref ref="sum-difference-derivative-rule" text="title"/>,
     <xref ref="constant-multiple-derivative-rule" text="title"/>,
-    <xref ref="thm_PowerRule" text="title">Power Rule</xref>,
+    <xref ref="thm_PowerRule" text="title"/>,
```
sean-fitzpatrick, line 700: fixed in my `main` branch

ptx/sec_deriv_prodquot.ptx
```diff
@@ -711,1446 +697,1395 @@
...
-                <m><var name="$f"/>'(<var name="$x"/>) = </m><var name="$deriv" width="15"/>
+                <var name="$df" width="80"/>
               </p>
             </statement>
-        </webwork>
-      </exercise>
+          </webwork>
+        </exercise>
 
-    <exercise>
```
sean-fitzpatrick, line 1490: Did you remove these two exercises, or add ww code and move them? (Sorry, hard to tell in the diff.)
Where there are ones like this without ww code, it's most likely these are new questions that Greg added in APEX V4 (after the mbx conversion). I put them in the right place, but didn't have time to code them as ww problems.

sean-fitzpatrick:
Ahh, sorry -- I suspect looking at this now that I should have chosen "start a review" instead of just adding comments.
In any case, a couple of minor things to double-check.

Alex-Jordan:
I just force pushed after breaking up the own "webwork revisions" commit into an actual "webwork revisions" commit and the rest that was file formatting and minor editing. I had forgotten that was in there (from spring editing) and I added the entire file on the first go around. Comments on your comments to follow.

ptx/sec_deriv_prodquot.ptx
Alex-Jordan: You're right that it can't go here. Schema validation confirms that. Formerly it was inside one of the list items, which seems logically incorrect given that it is the marker for (all) trig derivatives. I moved it to follow the theorem's title, and that passes validation. I will commit and force push after reviewing the other comments.

ptx/sec_deriv_prodquot.ptx
Alex-Jordan: Maybe missed because I asked you to avoid chapters 1--3? I will remove it, and include that in the next force push.

ptx/sec_deriv_prodquot.ptx
Alex-Jordan: The issue is that the other list item in this has a `p`. If one has a `p` and the other does not, I saw something bad wrt formatting in output. Can't recall if it was HTML or PDF, but somewhere the `p` induces some extra vertical spacing, and it looked off if one item had that but not the other. Maybe the other item doesn't need a `p`. If it is important to change this, we should look at the output in both formats before committing.

ptx/sec_deriv_prodquot.ptx
Alex-Jordan: I think this one is the WW problem at the new line 1351.

Alex-Jordan:
OK, I made those edits, backed up, re-committed, and force pushed.

ptx/sec_deriv_prodquot.ptx
sean-fitzpatrick: Ah, that could be it.
I think that's a known problem but can't recall if it's been brought up with Rob. (I think yes.) Without the `p` there sometimes isn't enough vertical space.

This makes sense then

PR #88, opened 2020-09-12
webwork revisions

Alex-Jordan:
WW for section 2.6.

It's hard to code these. In `x^2 + y^2 = 1`, you have `dy/dx = -x/y`, so the answer is a function in two variables (`-x/y`). That part is the easy part. But the answer is also `-(1-y^2)/(xy)`, which is equivalent for `(x,y)` on the circle. So something fancy needs to be done. I'd already done that part, but then adding in randomization complicates it all. And after I added randomization, I realized it was all for nothing! Most of these problems need to not be randomized, because most of them require some amount of manual grading (so they shouldn't vary from student to student after all). They need manual grading if you want to see the student actually using implicit differentiation instead of, say, solving for `y` first.

### DIFF
--- a/ptx/sec_deriv_basic_rules.ptx
+++ b/ptx/sec_deriv_basic_rules.ptx
@@ -1934,7 +1934,7 @@
             $dfx0 = $df->eval(x=>$x0);
             $fx0 = $f->eval(x=>$x0);
             $answer = $fx0 + $dx*$dfx0;
-            $l = Formula("$dfx0(x-$x0)+$fx0")
+            $l = Formula("$dfx0(x-$x0)+$fx0");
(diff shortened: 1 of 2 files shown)


sean-fitzpatrick:
I'm fine with no randomization on these. I don't recall if I've seen many good randomized problems on implicit differentiation.
Usually some unimportant detail is random, like the constant for a level curve.

At the end of the week I'm hoping to have time to get these changes into production.
I can try cherry-picking these commits over on my fork/branch (which is currently several dozen commits ahead of `master`) or I can make the same edits (via good ol' manual copy-paste) as new commits.
I think it's better practice to cherry pick but I've been wrong before :-)

PR #91, opened 2020-09-16
WW revisions to 3.2

Alex-Jordan:
With this, I've completed WW revisions for chapters 1--3.

Once you merge this, I would like to begin the process of making a new branch `main` on the `APEXCalculus/APEXCalculusPTX` repository. It will be this `master`, with certain things from your (Sean's) branches. And then after that, let's work out a workflow that has no more forking.

### DIFF
--- a/ptx/sec_graph_extreme_values.ptx
+++ b/ptx/sec_graph_extreme_values.ptx
@@ -1489,8 +1489,8 @@
               $high = int(-$b/2) + $h;
               $f = Formula("x^2+$b x+$c")->reduce;
               Context("Fraction");
-              $max = Fraction(max($f->eval(x=>$low),$f->eval(x=>$high),$f->eval(x=>$h)));
-              $min = Fraction(min($f->eval(x=>$low),$f->eval(x=>$high),$f->eval(x=>$h)));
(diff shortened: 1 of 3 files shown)


sean-fitzpatrick:
Great, thanks!
I think I can get these merged into my fork with some cherry picks (and maybe a bit of conflict resolution).

Would the path of least resistance here be to merge things into my fork, and then identify the things you don't want in the official repo?

Alex-Jordan:
I'm not sure ye what the path of least resistance will be. But I have your repo in my remotes, and I can play around with all the options on my end without you needing to do any merging. Maybe the one thing that would help on your end while I try this today would be if you pause on any development. Is that possible for today? No problem if not. The next day I could give this some time is Friday. And then there's always next week.

sean-fitzpatrick:
I am completely buried in teaching until at least Friday. And I have kids at home again because a runny nose means no school until COVID testing comes back clean.
So no risk of me doing anything :-)

Lots of little commits this month as I corrected flaws revealed by the LaTeX build. But that's done now.
(We need new styling for assemblage. But that doesn't touch the book.)

PR #92, opened 2020-09-23
fix PG infinite loop bug

Alex-Jordan:
This problem was entering an infinite loop if a student entered "power rule" or "power" as the answer.


sean-fitzpatrick:
Good catch!

PR #96, opened 2020-09-28
Docinfo fix

sean-fitzpatrick:
Moves the correction for the TikZ circles bug out of `docinfo/macros` and into `latex-image-preamble`.

Also, change `\DeclareMathOperator` to `\newcommand` using `\operatorname`.
Does not seem to have any effect on HTML build, and images all built correctly.


sean-fitzpatrick:
Hmm. More commits in this PR than expected. 3 out of 5 commits are upstream merge commits. Maybe just cherry-pick d74cbcc?

Also I probably shouldn't have my local files in the `.gitignore`. Probably there is some way to do this locally. Will investigate.

sean-fitzpatrick:
Figured out where things should go for the files I need for localized versions that I don't want tracked by git (because you're not going to want them in the main repo), so I'm not cluttering up the `.gitignore` for everyone.
Removed all the old LaTeX file references while I was at it.

Alex-Jordan:
No more cherry-picking :) If we do things right from here on out, we shouldn't need it, and I think cherry-picking could become the source of future git tangles.

The files changed looks good, so I will merge.

With the things in `.gitignore`, your scripts could build some of them into the `output/` folder, which is ignored.

sean-fitzpatrick:
Turns out you can put local files in `.git/info/exclude`

Mostly these are modified chapter files and top level files, since -- for example -- we teach Taylor polynomials in Calc 1, so I need that section in the chapter on derivative applications, and not in the chapter on sequences and series.
They all live in `ptx/`

Issue #97, opened 2020-10-15
WeBWorK errors in Section 2.3


sean-fitzpatrick:
These might be easy to fix but I'm limited on time, so recording things here for attention later (by myself, or @Alex-Jordan ).

In [Section 2.3](https://opentext.uleth.ca/apex-standard/sec_basic_diff_rules.html#subexercises-23) there are several exercises where the answers are wrong. Most seem to involve functions of theta.
Problems 14, 30, and 31 all return incorrect answers.

sean-fitzpatrick:
29 is also not working right. And apparently some people can see the function for 14 as 9sin(theta)+10cos(theta) (which corresponds to the answer given) but others (including me, on both Chrome and Firefox for Windows and Linux) are seeing 9sin(theta)+theta.

sean-fitzpatrick:
Also a sign error in 13: should be -3/8 t^2 for the second term.

sean-fitzpatrick:
Ah, no -- I see it now! There is a missing space: `$bcos(theta)` should be `$b cos(theta)`

sean-fitzpatrick:
Pretty sure this issue can be closed. Re-open if I'm wrong on that.

PR #98, opened 2020-10-16
corrections to some WeBWorK code

Alex-Jordan:
Here are some corrections to those WW problems from #97, plus a few more.


sean-fitzpatrick:
Thanks! I'll put my laptop to work on that overnight and try to have new HTML up this weekend.

Alex-Jordan:
OK. If you'd like guidance on just editing the HTML and base64 for directly, we could look at that tomorrow in the PTX meeting.

sean-fitzpatrick:
Thanks. I'll try to remember to drop by. I don't have teaching Friday afternoons but I do have a kid at home. Our school system only does half days on Friday.

PR #99, opened 2020-10-16
webwork coding error

Alex-Jordan:
Hope you didn't start that build yet, because here is a correction to one of my corrections.

### DIFF
--- a/ptx/sec_deriv_basic_rules.ptx
+++ b/ptx/sec_deriv_basic_rules.ptx
@@ -1051,7 +1051,7 @@
               Context("Fraction");
               Context()->variables->are(t=>'Real');
               $f=Formula("$a t^5 + 1/$b t^3 + $c t + $d")->reduce;
-              $fp=Formula("5 $a t^4 + 3/$b t^2 + $c")->reduce;
+              $fp=Formula("5*$a t^4 + 3/$b t^2 + $c")->reduce;
             </pg-code>
             <statement>
               <p>
--- a/ptx/sec_deriv_chainrule.ptx
+++ b/ptx/sec_deriv_chainrule.ptx
@@ -1958,8 +1958,8 @@
             $f = Formula("$a x^$n e^x $trig(x)")->reduce;
             $g = Formula("$b^x $c^x $d^x");
             Context()->flags->set(reduceConstantFunctions=>0);
-            $fp=$f->D($x);
-            $gp=$g->D($x);
+            $fp=$f->D('x');
+            $gp=$g->D('x');
           </pg-code>
           <statement>
             <p>


sean-fitzpatrick:
I did, but it's only my computer that had to do extra work. (It very nearly didn't survive the night: a leak in our dishwasher came through the ceiling to my basement office and missed my laptop by inches -- it was covered in splatter marks from the drops splashing onto my desk!)

One more was spotted by a student today: question 44 on the chain rule section has the wrong answer. (The derivative of 2^x3^x4^x is not 0)

sean-fitzpatrick:
I've looked through the code for the last problem in the chain rule section, and compared it to some earlier ones and I can't figure out where it's going wrong. But the answer for both derivatives is coming out as 0. Here's the code:
```
<pg-code>
            Context()->flags->set(reduceConstants=>0);
            $a=random(1,9,1);
            $n=random(1,4,1);
            $trig=list_random('sin','cos','sec','csc','tan','cot');
            ($b,$c,$d)=num_sort((2..9)[NchooseK(8,3)]);
            if($envir{problemSeed}==1){$a=2;$n=1;$trig='cot';$b=2;$c=3;$d=4;};
            $f = Formula("$a x^$n e^x $trig(x)")->reduce;
            $g = Formula("$b^x $c^x $d^x");
            Context()->flags->set(reduceConstantFunctions=>0);
            $fp=$f->D($x);
            $gp=$g->D($x);
          </pg-code>
```

Alex-Jordan:
Sorry, I thought I fixed this the other day. I think I must have been in the middle of doing it and got pulled away by the kids or something. But it's because of `$x` that should be `'x'`.

sean-fitzpatrick:
Thanks. Turns out I didn't make the drop-in meeting Friday because our dishwasher decided to send water through the ceiling of my office in the basement. Missed my computer by about 3 inches.

sean-fitzpatrick:
I've rebuild and update the HTML. Looks good.

How hard is it to get WeBWorK to display an answer like `\ln(2)2^x` as such, rather than as `0.69315 2^x`?

Alex-Jordan:
It's not hard. The issue becomes that you can make "ln(2)" stay as "ln(2)" but then also "ln(e)" will stay as "ln(e)" and "ln(e^5)" will stay as "ln(e^5)". So if it's a situation where both kinds of thing can show up depending on randomization, you either live with one of them being imperfect or you make tedious conditional code.

Add the line `Context()->flags->set(reduceConstantFunctions=>0;` following a context declaration (or near the top of the code portion if there is no context declaration) and then functions being applied to constants (like ln(2)) will not actually be evaluated.

sean-fitzpatrick:
Ahh... yes, I remember dealing with that at some point now. (Or at least reading about it on the WW forums.)

Maybe I'll poll my class and see if they have a preference.

sean-fitzpatrick:
In case you get to it before me (likely, because I teach in 15 minutes, for class 2 of 3 on the day).

This is problem 44 in Section 2.4 (sec_deriv_prodquot.ptx, starting at line 1825):
```
        <exercise>
          <webwork seed="1">
            <pg-code>
              Context("Fraction");
              Context()->noreduce('(-x)+y','(-x)-y');
              ($a,$b,$c) = (-3..-1,1..3)[NchooseK(6,3)];
              if($envir{problemSeed}==1){$a=1;$b=1;$c=1;};
              $formula = Formula("($a x^2)/($b x + $c)")->reduce;
              @x = (Fraction(0),Fraction(-$b/(2*$c)));
              $answer = List(@x);
            </pg-code>
            <statement>
              <p>
                <m>f(x) = <var name="$formula"/></m>
              </p>
              <instruction>
                Use commas to separate more than one <m>x</m>-value.
                If there are no such <m>x</m>-values,
                answer with <c>NONE</c>.
              </instruction>
              <p>
                <var name="$points" width="15"/>
              </p>
            </statement>
          </webwork>
        </exercise>
```
The answer comes out as 15; should be the list `0,-2`. Two mistakes, I think:
1. `$answer` should be `$points` (or vice-versa)
2. The second point on the list is given as `Fraction(-$b/(2*$c)` but should be `Fraction(-(2*$c)/$b)`.

sean-fitzpatrick:
Question 40 (same section, starting at line 1699) also does not give the correct answer:
```
        <exercise>
          <webwork seed="1">
            <pg-code>
              ($a,$b) = (-9..-1,1..9)[NchooseK(18,2)];
              $trig = list_random('cos','sin');
              if($envir{problemSeed}==1){$a=-8;$b=1;$trig='cos'};
              Context("Fraction");
              $formula = Formula("($trig(x)+$a x)/(x+$b)")->reduce;
              $x0=0;
              $y0=Fraction($formula->eval(x=>$x0));
              $df=$formula->D('x');
              $m=Fraction($df->eval(x=>$x0));
              Context()->variables->add(y=>'Real');
              parser::Assignment->Allow;
              $t=Formula("y=$m +$y0")->reduce;
              if ($m == 0) {
                $n=Formula("x=$x0");
              }
              else {
                $mn = Fraction(-1/$m);
                $n=Formula("y=$mn +$y0")->reduce;
              }
            </pg-code>
            <statement>
              <p>
                <m>g(x) = <var name="$formula"/></m> at <m>\left(<var name="$x0"/>,<var name="$y0"/>\right)</m>
              </p>
              <instruction>
                Enter the equation of the tangent line here.
              </instruction>
              <p>
                <var name="$t" width="20"/>
              </p>
              <instruction>
                Enter the equation of the normal line here.
              </instruction>
              <p>
                <var name="$n" width="20"/>
              </p>
            </statement>
          </webwork>
        </exercise>
```

sean-fitzpatrick:
Q40 is missing (x-$x0) in the def of $t and $n. Fixing now

PR #101, opened 2020-10-22
webwork coding errors

Alex-Jordan:
These two problems had two things wrong with them. One is that instead of something like (x - pi/2) in the formula, it was just (x - 2). The other is that the normal slope was not taking the reciprocal of both of its product components.

### DIFF
--- a/ptx/sec_deriv_basic_rules.ptx
+++ b/ptx/sec_deriv_basic_rules.ptx
@@ -1792,14 +1792,14 @@
               $m = Formula("-$b")->reduce*$trig{$trigd{$c}}{$a} if ($c eq 'cos');
               Context()->variables->add(y=>'Real');
               parser::Assignment->Allow;
-              $t=Formula("y=$m(x-$a)+$k");
+              $t=Formula("y=$m(x-pi/$a)+$k");
               $t=Formula("y=$k") if ($m == 0);
               if (($c eq 'sin') and ($a==2))
                 {$n= Formula("x=pi/$a");}
               else {
-                $rm = Formula("$b")*$trigrecip{$trigd{$c}}{$a};
-                $rm = Formula("-$b")->reduce*$trigrecip{$trigd{$c}}{$a} if ($c eq 'cos');
-                $n= Formula("y=-$rm(x-$a)+$k");
+                $rm = Formula("1/$b")*$trigrecip{$trigd{$c}}{$a};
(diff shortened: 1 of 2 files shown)


sean-fitzpatrick:
Here is another problem with a mistake. Since you have a PR open, do you want to add it?

Section 3.1, 18. Begins at line 1546 of `sec_graph_extreme_values.ptx`.
Here is the code:
```
<exercise>
          <webwork seed="1">
            <pg-code>
              Context("Fraction");
              $r = random(-9,-1,1);
              $s = random(1,9,1);
              $d = non_zero_random(-9,9,1);
              $high = $s+random(1,2,1);
              if($envir{problemSeed}==1){$r=-2;$s=5;$d=3;$high=6};
              $low = Real(0);
              $b = Fraction(-3*($r+$s)/2);
              $c = 3*$r*$s;
              $f = Formula("x^3+$b x^2+$c x+$d")->reduce;
              $max = Fraction(max($f->eval(x=>$low),$f->eval(x=>$high),$f->eval(x=>$r),$f->eval(x=>$s)));
              $min = Fraction(min($f->eval(x=>$low),$f->eval(x=>$high),$f->eval(x=>$r),$f->eval(x=>$s)));
            </pg-code>
            <statement>
              <p>
                <m>f(x) = <var name="$f"/></m> on <m>[<var name="$low"/>,<var name="$high"/>]</m>
              </p>
              <instruction>
                Enter the maximimum value here, or enter <c>NONE</c> if there isn't one.
              </instruction>
              <p>
                <var name="$max" width="10"/>
              </p>
              <instruction>
                Enter the minimimum value here, or enter <c>NONE</c> if there isn't one.
              </instruction>
              <p>
                <var name="$min" width="10"/>
              </p>
            </statement>
          </webwork>
        </exercise>
```
Looks like the max is evaluated as f(-2), but the interval is given as [0, 6].

Maybe the problem is defining `$low=Real(0);` as the left endpoint of the interval, rather than basing it on `$r,$s,$d`?

sean-fitzpatrick:
Looks good. Do you want me to leave this open in case we spot a few more, or merge for now?

Alex-Jordan:
I don't mind opening new PRs as the need arises. So you could merge.

There are two instructors here trying out the problems in live WW sets. They bring errors to me now and then. So their sections plus yours are vetting things right now.

sean-fitzpatrick:
OK. Merging now.

PR #102, opened 2020-11-05
Fix typo in T/F question

sean-fitzpatrick:


Alex-Jordan:
Catching up on several weeks of email. I think this is not right, because shouldn't it just be `dy`, not `\dy`?

Alex-Jordan:
Oh, I see now. `\dy` is a macro for Delta y. Sorry!

sean-fitzpatrick:
Yeah. This one was just a quick fix so I merged it myself. The question was written twice and one of the two didn't use math tags.

PR #103, opened 2020-11-12
Update Asymptote code to allow black and white version

sean-fitzpatrick:
This PR replaces all the RGB colour definitions with named pens. Those pen names are then assigned values in an `asymptote-preamble` section in the `apex.ptx` and `apex-print.ptx` top level files.

Not yet tested -- I need to update PreTeXt first to write the `asymptote-preamble` in during asymptote extraction.
That's next on my list.

But if you spot any obvious errors, please let me know.


sean-fitzpatrick:
This is mostly working now. I have a few asymptote figures that don't compile. But the issue is a Ghostscript error related to opacity settings in some labels. Online forums suggest adding a `-nosafe` option to the `asy` executable but this didn't work for me.

sean-fitzpatrick:
I think this is now working as expected. A few figures look a little bit different because we did things like use `surfacepen` to replace `rgb(0,0,.7)` with different opacity settings (everything from 0.4 to 0.8 is now 0.7, I think. The 0.4 was in an exercise and a bit of an outlier.)

Alex-Jordan:
This looks like it's all asymptote related in the latter part of the book. I'm pretty sure I won't get around to testing anything out, so I'll just merge this based on your testing.

PR #108, opened 2020-12-27
Holiday editing

sean-fitzpatrick:
Mostly adding subexercises in the remaining chapters, but I also removed some redundant asymptote code and made a few other small changes.

### DIFF
--- a/ptx/chapter_deriv_apps.ptx
+++ b/ptx/chapter_deriv_apps.ptx
@@ -14,8 +14,6 @@
   <xi:include  href="sec_related_rates.ptx"/>
   <xi:include  href="sec_optimization.ptx"/>
   <xi:include  href="sec_differentials.ptx"/>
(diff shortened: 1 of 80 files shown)


Alex-Jordan:
Are you sure about removing the `sidebysides`? It sounded like PTX was going to change to allow natural width for `tabular`s.

sean-fitzpatrick:
Do you mean the changes in `4f1e960`? All of those were sidebysides where one of the two panels was a paragraph and the other was a tabular or figure. I just didn't like how they looked in print, and realized that with our PDF styling, the tabular/image would migrate to the margin.

They didn't look particularly great in HTML either.

Alex-Jordan:
OK, I had forgotten about the margin styling.

There's too much for me to actually look it all over with a closer eye, so I'm just going to merge.

PR #109, opened 2020-12-30
Add 'instructions' to later WW exercises, and some other small edits

sean-fitzpatrick:
In later chapters there are a number of sections where roughly half the exercises are coded for WeBWorK, and half are not.
The two do not look good side by side in print.

I've tried to streamline things so that everything comes out the same. (Building LaTeX right now, so wait for confirmation tomorrow.)

There were also a few minor things I caught while proofreading the PDF.

### DIFF
--- a/ptx/sec_antider.ptx
+++ b/ptx/sec_antider.ptx
@@ -126,7 +126,7 @@
 
       <!-- Todo: Fix this weird image -->
 
-      <image width="100%">
+      <image xml:id="img_anti1" width="100%">
         <description></description>
         <latex-image>
 
--- a/ptx/appendix_back_reference.ptx
+++ b/ptx/appendix_back_reference.ptx
@@ -190,7 +190,7 @@
 
     <list xml:id="reference-int-elementary">
       <title>Integrals of Elementary (non-Trig) Functions</title>
-      <ol>
+      <ol cols="2">
         <li>
           <m>\int x^n\,dx=\frac{1}{n+1}x^{n+1}+C, n\neq -1</m>
         </li>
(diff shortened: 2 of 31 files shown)


ptx/sec_antider.ptx
```diff
@@ -126,7 +126,7 @@
 
       <!-- Todo: Fix this weird image -->
 
-      <image width="100%">
+      <image xml:id="img_anti1" width="100%">
```
Alex-Jordan, line 129: I maybe forgot to alert you, but I'm working on chapter 5 now. So this may lead to a conflict. No big deal, but you could hold any more edits to the exercises in ch 5.

Alex-Jordan:
I only looked closely at code changes in chapters 1--5. Looks good.

ptx/sec_antider.ptx
sean-fitzpatrick: There were about a dozen images that didn't have `xml:id`s. That's the change here.

sean-fitzpatrick: That also is the only change to Chapter 5. (didn't realize you were working on it)

sean-fitzpatrick:
The latex file is ready to go but I need to merge in all the hand edits. And right now I'm in the middle of an online tasting of imperial stouts so this might not be the best time to do textbook work.

sean-fitzpatrick:
Caught two dumb mistakes

Alex-Jordan:
6:07: Sean is tasting imperial stouts.
6:52: Sean makes two commits :)

I meant to merge this with my last comment. But I was on hold on the phone and that's when I got through.

sean-fitzpatrick:
Good thing they kept you on hold, I guess. I can't even blame the beer for the mistakes I caught.

PR #111, opened 2021-01-06
Chapter 5 WW revisions

Alex-Jordan:
Actually this skips section 5.5 for now, because the local instructor covers that at the end of the term. I'm moving on to chapter 6 now.

I just noticed this unintentionally committed `Makefile` and `apex.ptx`. I'll undo that and force push. Probably will be addressed before you even see this.


Alex-Jordan:
Oops, messed that up. I'll start over.

PR #120, opened 2021-02-08
Main

Alex-Jordan:
WW revisions for 6.2, 6.3. And some corrections to WW in 6.1.

### DIFF
--- a/ptx/sec_IBP.ptx
+++ b/ptx/sec_IBP.ptx
@@ -211,20 +211,20 @@
   </p>
 
   <p>
-    A useful mnemonic for helping to determine <m>u</m> is <q>LIATE,</q> where
+    A useful mnemonic for helping to determine <m>u</m> is <q><acro>LIATE</acro>,</q> where
   </p>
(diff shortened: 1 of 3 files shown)


sean-fitzpatrick:
Looks good, though GitHub didn't give me a great diff for the bigger changes.

How is `<acro>` different from `<init>`? (I knew we had `<init>` but didn't know we had `<acro>`.)

Alex-Jordan:
> How is <acro> different from <init>? (I knew we had <init> but didn't know we had <acro>.)

I don't know if they are functionally different right now in any ouput form. But an acronym would be spoken as a word, like SCUBA. And with initialism, you say each letter. Like USB. (And then there are abbreviations, where you are supposed to say the whole word that it stands for, even though you only see something like MON.)

sean-fitzpatrick:
OK, sounds good! There's a chance that at some point I've used `<init>` where I should have used `<acro>`. But most things are actually initialisms, like FTC.

PR #121, opened 2021-02-16
move knowl options to publisher file

sean-fitzpatrick:
What it says on the label.


sean-fitzpatrick:
There are now some additional error correction changes here, because I neglected to make a separate branch.
I'm going to merge these so I can do the same for the WW fixes.

PR #122, opened 2021-02-22
Figure incr1line missing annotations.

grady:
The number line for the first example in Increasing-Decreasing functions is missing annotations shown in the print version.

### DIFF
--- a/ptx/sec_graph_incr_decr.ptx
+++ b/ptx/sec_graph_incr_decr.ptx
@@ -330,6 +330,9 @@
                 ]
                 \addplot[guideline] coordinates {(-1,0) (-1,2)};
                 \addplot[guideline] coordinates {(0.3333,0) (0.3333,2)};
+                \addplot[mark=none] coordinates {(-1.6666,1)} node {\parbox{3em}{\centering \small $f'>0$ incr }};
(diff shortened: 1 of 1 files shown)


sean-fitzpatrick:
I think the annotations you want are in the second figure in that example. (There used to be only one figure -- now there are two.)

The undecorated version is an addition that wasn't in the print version. (@Alex-Jordan maybe this was done by you or someone else in the PCC group during the first attempt at conversion?)

That said, the diagrams need work -- they don't look as nice as they did in print. Might be a case of just needing to recalculate dimensions. Earlier in the conversion process someone had replaced the TikZ images with a table, I think probably because the TikZ wasn't showing nicely.

I *think* -- but could be wrong -- that the thinking here was to first make the number line, then explain what to do with it, and finally fill it in.

grady:
Oops, looks like you're right, missed that second figure with everything filled in.

Alex-Jordan:
Looking nice is important. Accessible images is also important. Putting the annotations in the `latex-image` means certain people can't read them. I think that might have led to a table version of this. 

Can't look deeper into it right now, but that's some history that immediately comes to mind. Can keep discussing in coming days.

sean-fitzpatrick:
Haha... I think we are all to busy to look at this right now. Accessible sign diagrams that look nice would be great. I didn't really like the table look, but I admit that I didn't think of the accessibility issue at the time. The fact that we don't have descriptions on images doesn't help, either.

PR #125, opened 2021-03-17
Corrected annotation in fig_anti1

grady:
In Section 5.1, figure 5.1.5 (integral notation), the integrand function is mislabeled as "Integration Symbol" at line 134. 

I also made a couple other minor changes to the annotation text for clarity and to suit my preferred style, see the diff for details.

### DIFF
--- a/ptx/sec_antider.ptx
+++ b/ptx/sec_antider.ptx
@@ -127,11 +127,11 @@
 
         \node[] (integral)
         {$\displaystyle\int$};
-        \node[above left of=integral, anchor=south east] (l1) {Integration Symbol};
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Looks legit :)

If it were just the first tow things, I would merge right now. With the other two things, would @APEXCalculus mind taking a look?

APEXCalculus:
All changes are fine with me.

Changes 1, 2 & 4 are an easy "yes". Change #3 from "One antiderivative" to
"Any antiderivative of $f$" is fine. At first, I thought it neither made
the graphic better nor worse; more thought makes me it makes the graphic
more descriptive.

Thanks.

PR #129, opened 2021-03-22
Fix typos

sean-fitzpatrick:
Sorry for messing up my commit history -- I had some uncommitted typo fixes, and there were upstream changes I hadn't merged.

This *should* only bring in some minor changes: the residual LaTeX noticed by @grady as well as some that I noticed later on. (Greg had some macros like `\vrp` that used to use `\ensuremath` so that they worked in text mode. I caught most of them but missed a few.)


ptx/sec_taylor_poly.ptx
```diff
@@ -583,16 +583,16 @@
...
+                \addplot+ [rightarrow,domain=0.01:3,smooth,samples=50] {ln(x)} node [pos=1,above left] { $y=ln(x)$};
+                \addplot+ [rightarrow,domain=-4:4,smooth] coordinates {(-0.108,-7.567)(-0.052,-4.958)(0.004,-3.519)(0.06,-2.671)(0.116,-2.13)
+                (0.172,-1.756)(0.228,-1.478)(0.284,-1.259)(0.34,-1.079)(0.396,-0.9263)
+                (0.452,-0.7941)(0.508,-0.6773)(0.564,-0.5727)(0.62,-0.478)(0.676,-0.
+                3916)(0.732,-0.312)(0.788,-0.2383)(0.844,-0.1696)(0.9,-0.1054)(0.956,-
+                0.045)(1.012,0.01193)(1.068,0.06579)(1.124,0.1169)(1.18,0.1655)(1.236,
+                0.2119)(1.292,0.2562)(1.348,0.2986)(1.404,0.3393)(1.46,0.3784)(1.516,
+                0.4161)(1.572,0.4523)(1.628,0.4874)(1.684,0.5212)(1.74,0.5538)(1.796,
+                0.5853)(1.852,0.6154)(1.908,0.6427)(1.964,0.6635)(2.02,0.6665)(2.076,
+                0.621)(2.132,0.4476)(2.188,-0.04877)(2.244,-1.326)(2.3,-4.421)} node [pos=0.8,below right] { $y=p_{20}(x)$};
```
Alex-Jordan, line 595: Where possible, I like to replace these point lists with more meaningful code expressions. Here, like `-(x-1)^20/20+(x-1)^19/19±...`. For one thing, the error here would have been caught sooner.

Using `smooth` raises some flags. It's probably fine, but since it occasionally does bad things, I try to avoid it. Did you try just upping the `samples`?

Only if you are interested...

For `ln(x)`, plotting it as a function of x near 0 of course causes smoothness issues. In situations like that, using parametric plotting and transformations makes good results. Like, try this:
```
\addplot[domain=0.01:3] ({x},{ln(x)});
```
which is bad. But then try this:
```
\addplot[domain=-4.6:1.1] ({e^x},{x});   %y=ln(x)
```
or perhaps better:
```
\addplot[domain=-4.6:1.1,variable=t] ({e^t},{t});   %y=ln(x)
```
No need for `smooth`, or boosting `samples` past the default 20.

sean-fitzpatrick:
@Alex-Jordan please feel free to change this. (Or send me the changes.) This is for the figure @grady pointed out, where the graph for `sqrt(x)` was being used instead of `ln(x)`. I think this mistake might date back quite a ways, and never got spotted.

I tracked down the original TikZ code from @APEXCalculus and dropped it in, with some adjustments for the global styling changes. Quick fix to at least have the right graph in there. `smooth` was in Greg's code (which I think used `smooth, thick` throughout) and can probably be dropped.

sean-fitzpatrick:
domain needs adjusting as well

Issue #128, opened 2021-03-22
Problem with Taylor Poly ln(x) figure fig_taypoly2c


grady:
It appears that the figure fig_taypoly2c, supposedly showing the 20th-degree polynomial of ln(x) is incorrect. It seems to be identical to the figure for the 4-th degree expansion of sqrt(x) fig_taypoly5c shown later in the section.

Also, in fig_taypoly5c at line 1131 the function is labeled ln(x) when it should be sqrt(x).

sean-fitzpatrick:
Yes, I see that too. Good catch. I *think* I might have time to look at this tomorrow.
But I've frequently been wrong about what I have time for this semester. :-)

sean-fitzpatrick:
Also there's an issue with how that figure is cropped: the label gets cut off. Not sure if that's happening at the cropping stage, or at the TikZ stage.

sean-fitzpatrick:
This should be fixed once #134 is merged.

PR #134, opened 2021-06-26
Re-index partitions for Riemann sums

sean-fitzpatrick:
I'm pretty sure I've caught all cases of partitions in the book, and changed the indexing so that we go from 0 to n instead of 1 to n+1.
Let me know if you find any screw-ups. I'll double check this in about 10 days. Going camping for a week starting tomorrow, so I'll be offline (mostly).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ output/
 webwork/
 asy/
 #
+# Macro file for WeBWorK
+APEX.pl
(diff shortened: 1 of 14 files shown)


sean-fitzpatrick:
Last commit doesn't fit the theme of this PR but I didn't want to forget to do it.
(It seems that I fixed the bigger issue in #128 (the wrong image was being used in `fig_taypoly2c`) but forgot to correct the labels)

sean-fitzpatrick:
I think this is safe to merge now (and I'd like to do so soon to avoid merge conflicts with other upcoming work).
@APEXCalculus let me know if you see anything out of place.

APEXCalculus:
Took me a while to read through each change - figured Sean did all the work so a fresh set of thorough eyes would be good, but sorry for the length of time it took.

PR #136, opened 2021-07-16
Updates and edits 2021

sean-fitzpatrick:
Further work to come, so this should be left open for now.

Updates so far:
- Adds example and discussion for implicit differentiation with a function of three variables
- Changes spherical coordinate convention to use elevation angle


sean-fitzpatrick:
Also various typo fixes found along the way.

APEXCalculus (approved):

PR #137, opened 2021-07-19
Publication updates

sean-fitzpatrick:
This pull request updates the LaTeX style sheets to use code for margin figures and margin notes based on Greg's original LaTeX.
There is also an update to move another string parameter into the publisher file.

Some notes:

- right now I've coded things differently for one-sided (all figures in right margin) vs. two-sided (margin alternates left and right). I can combine both together with an `xsl:choose` conditioned on `@latex-sides` but this adds a lot of bloat to the code, and we have separate style files for print and electronic anyway.
- the vertical placement is hard-coded to default to `0.5\pageheight`, and once the `.tex` file is built, each figure will have to be manually adjusted

Suggestions welcome on: whether to add the `xsl:choose` so that things work if someone uses the wrong setting in the publisher file, and whether we want to try to build hints for the vertical placement into the `ptx` source somehow.


sean-fitzpatrick:
Working on a complete refactoring of this, using `tcolorbox`. Changes coming soon!

sean-fitzpatrick:
I think this is ready to go. This implements the `xcoffins` solution provided [on TeX Stack Exchange](https://tex.stackexchange.com/questions/605955/can-i-avoid-indentation-of-margin-items-when-using-parbox-false-in-a-tcolorbox?noredirect=1#comment1519900_605955). I currently have an older version that uses hard-coded lengths. I might try his updated version, which calculates the hshift automatically, but the current one works.

@Alex-Jordan let me know if you prefer another method for dealing with print/colour versions, or if you want different organization. I've made two versions of both the publisher file and the docinfo file, and placed those in folders within `style/`. The bash scripts I use to build the book now start by copying the appropriate version in. Eventually the `docinfo` can be handled with a `@version` specification in the publisher file, but this isn't yet publicly available in PreTeXt.

sean-fitzpatrick:
Closing this PR since all the commits here are included in #139. But feedback on the reorganization is still welcome.

PR #145, opened 2021-09-21
Fixed simple typo

APEXCalculus:
fixed a small typo from -pi to -\pi.


sean-fitzpatrick:
Ha! Thanks for that. It was on my to-do list.

I found it amusing that the person sending the email felt it necessary to explain how to fix it.

APEXCalculus:
I didn't think about the suggestion - that is funny. "I know you've coded
thousands of pages of math, but here's a suggestion for making the pi
symbol."

Issue #141, opened 2021-09-02
Problem 1.3.15


grady:
In the analytical limits section there is a problem that asks for a limit of f(x)^{g(x)}. Problem number 15 in my build, and the limit ends up being 2^0 = 1. The issue is that we don't have any limit property in the section that directly applies to this situation. The expression is neither a power expression nor an exponential expression, because both the base and exponent are functions.

This problem is answerable using the limit properties in the section, but it seems to require the use of the exp(log(...)) trick. Once you've algebraically rewritten the problem as equivalent to exp(g(x) * ln(f(x)) (which is only really known to be true in some neighborhood of 1, based on the info given), you then need to use the composition property, the continuity of log, the multiplication property, the composition rule again , and the continuity of exp in order to reach the correct conclusion about the limit. (Provided one uses only the results in the book so far.)  

The difficulty of this problem (if done correctly) seems way out of proportion to its place in the order of problems. For example the next three problems don't require any algebraic tricks to apply the limit properties, and only require the use of between 1 and 3 limit properties in order to reach the desired conclusion.

I like the problem, but it's a real challenge. I think it should be maybe moved closer to the end of the list of problems?

![Screen Shot 2021-09-01 at 16 53 22](https://user-images.githubusercontent.com/60004/131774048-1721124d-e1ca-4356-803e-a7a9c0ecd184.png)

sean-fitzpatrick:
I feel like this one came up before and we considered making the answer "not possible to know", given the techniques currently at hand.

@APEXCalculus what do you think? Make this a question on its own?

grady:
Yes, after reflecting on this I think that perhaps the original intention of the problem was that the answer was supposed to be "not possible to know" (unless a student is very clever and knew to try the exp-log thing), but at some point during the conversion to the WebWork, that the answer got "fixed" to be 1.

APEXCalculus:
Grady gives me too much credit. That problem was written while I was
thinking like a Civil Engineering Freshman, and not like the Professional
Mathematician that I'm paid to be. The WebWork conversion didn't make the
answer 1; Hartman did from the start.

Or we can blame Sean. I'm good with that, too.

That problem should be changed. I propose changing it to 5^f(x). The
nuances of why we can't answer the original problem are too complex to be
introduced, and resolved, through the Exercises.

grady:
I like the proposed replacement problem, since there is no other appearance of an exponential limit in the exercises so far. 

[...]

PR #146, opened 2021-09-24
Collected corrections

sean-fitzpatrick:
I've had a branch running since August containing several small fixes I caught while finalizing publishing for the fall.

Originally this was supposed to be where I would set up versioning in case someone wanted the good ol' spherical coordinates of APEX 4.0, but I ran out of time for that. Changes include:

- Add "old" versions of the two sections affected by the change in spherical coordinates. (The only other section that changes is the section I wrote on Jacobians.)
- Remove some of the redundant XSL style sheets. Especially for HTML, the only differences between the three that were there were which lines were commented out.
- Final changes to the PDF styling. (Word of caution: my current PDF versions rely on a slight tweak to `pretext-latex.xsl` to only display QR code (and not thumbnail) for videos. Maybe there's a way to override this in our XSL.)
- Fix two video codes that were accidentally swapped in the derivative intro section. This was noticed by Greg in a discussion with Grady about formally adding an example showing that every differentiable function is continuous.
- Fix an issue with a limit exercise noticed by Grady.
- Get the original Taylor polynomials section caught up with the version I'm using in my books. (This is a section I need to address with versioning so I don't have to maintain parallel versions.)

Let me know if you want me to cherry pick some of this commits into separate pull requests.


sean-fitzpatrick:
I am going to add a few more commits to this PR:
- Chuck Cusack has sent me a few typos that need fixing
- In #141 we have the discussion of renaming one of the two theorems named "Special Limits". I think I'll go with "Limits of Common Functions" rather than "Specialer Limits" or "Special Limits 2: Electric Boogaloo".

After that, someone should let me know if there's any reason not to merge this into `main`. Most of the commits are routine maintenance.

Alex-Jordan:
I combed through the minor changes. I defer to @sean-fitzpatrick on the larger changes here like with the Taylor polynomials sections and a few other things. Merging.

PR #154, opened 2022-01-14
Fix one typo and one formatting flaw

sean-fitzpatrick:
These are changes in a section that is not part of official APEX. It's a supplement I added for a handful of interested Calc 4 students. (So it is rarely viewed by anyone, but someone on the internet emailed to tell me about how I'm bad at LaTeX.)


Alex-Jordan:
Hey, I merged something :)

Issue #157, opened 2022-02-01
Small correction to Ex 6.5.9.


grady:
In [Example 6.5.9: Integrating using partial fractions](https://opentext.uleth.ca/apex-video/sec_partial_fraction.html#ex_pf5) when solving the linear system for the coefficients:

https://github.com/APEXCalculus/APEXCalculusPTX/blob/a04e5676a76fc6817ae085ebda996e8e4b0d1120/ptx/sec_partial_fraction.ptx#L531-L534

At first it is stated (correctly) that A=5, then later it becomes (wrongly) A=1 in L533

PR #158, opened 2022-02-01
Fix for issue #157.

grady:
Resolves issue #157.  I considered changing it to 11A=55, but this is the most straightforward correction.


sean-fitzpatrick:
Thanks Grady. Definitely looks like a simple one character typo. Merging.

PR #163, opened 2022-02-22
Removed '!'.

APEXCalculus:


Alex-Jordan:
I came to merge this, but then I had an observation. For reference, this can be seen by opening the solution here:

Example 2.1.24: Finding the derivative of a function 
https://opentext.uleth.ca/apex-video/sec_derivative.html#ex_deriv_sinx

We, the readers, are just as excited as the author to find this result. An alternative to removing the exclamation point is to move it over to the column that already has text. Possibly with some words too, like "Wow!" or "Remarkable!" or whatever. Are you interested in doing that instead?

APEXCalculus:
I like Alex's suggestion better of putting a note in the text column. See other pull request.

Issue #165, opened 2022-03-07
open and closed


Alex-Jordan:
An open interval has the form (a,b).

And a closed interval has the form [a,b]. Except that [a,∞) is also an interval that is closed.

Should we use vocabulary like "open intervals and compact intervals" instead of "open intervals and closed intervals"? I am thinking in the context of "when the domain of a function is a closed interval...". 

For example this theorem is not true, even if we all know what is meant:
Theorem 3.1.4: The Extreme Value Theorem 
https://opentext.uleth.ca/apex-video/sec_extreme_values.html#thm_extreme_val

Or does APEX define "closed interval" in a special way? I looked a bit but I'm not sure where that would be defined. There is a definition for "closed" in the multivariable content that would imply [a,∞) counts as closed, but we already acknowledge that some multivariable definitions to not reduce to the single variable definitions just because n=1.

I'm not opposed to defining "closed interval" to be limited to intervals of the form [a,b], for pedagogical reasons to avoid a new term like "compact". If it's not there yet, should we add it to section 3.1?

Another thing is to meticulously examine each instance of the phrase "closed interval" and see if it really should say "closed interval of the form [a,b]".

APEXCalculus:
Good catch.

Let's avoid new terminology like "compact". Here's a proposed solution to
the issue:

1. The first time we use the term "closed interval" - which I think is in
Section 1.5, Continuity - we make a marginal note that says something along
the lines of "In this text, when we use the term 'closed interval', we mean
an interval of the form [a,b], where a and b are real numbers. One may be
surprised to learn that intervals of the form [a,\infty), (-\infty, b] and
even (-\infty,\infty) are all also considered closed in advanced calculus.
While the mathematics supported by this definition of closed is
fascinating and important, it is beyond the scope of our purposes here."

2. Whenever we need a closed interval for a theorem, we explicitly use
[a,b] terminology. For instance, the MVT could read "Let f be a
continuous function defined on a closed interval I = [a,b]. Then f has both
a ..." Or eschew naming the interval and just use "Let f ... on a closed
interval [a,b]. Then f ... on [a,b]."

Thoughts?
Greg



(note to self: apex calculus error errata things to fix)

sean-fitzpatrick:
This would be an easy fix. I agree that "compact" risks overloading students who already have too much terminology to tackle.
It would not take long to search the whole book for all occurrences of "closed interval" and where necessary, replace with "closed interval [a ,b]".

[...]

Issue #156, opened 2022-01-24
small edit needed(?) for continuity


Alex-Jordan:
Theorem 1.5.12: Continuous Functions 
https://opentext.uleth.ca/apex-video/sec_continuity.html#thm_continuous_functions

says that certain functions are "continuous on their domains". For functions like tangent, the domain is neither an open interval nor a closed interval, which are the only types of set for which "continuous on a set" has been defined/discussed. Noting here to seek out a way to adjust the wording to make this fit well without becoming too verbose or technical.

sean-fitzpatrick:
Can we just add something like:
```
<aside>
  <p>
    When the domain of a function <m>f</m> is a disjoint union of open intervals,
    we say that <m>f</m> is continuous on its domain if it is continuous on each interval.
  </p>
</aside>
```
Maybe there could be a second sentence with an xref to the example of the floor function to urge caution in the case of half-open intervals.

APEXCalculus:
I like the basics of Sean's suggestion. How about relating it directly to the Theorem at hand, with the following adjustment:

<aside>
  <p>
    For functions <m>f</m> (such as <m>\tan x</m>, <m>\cot x</m>, <m>\sec x</m> and <m>\csc x</m>)  where the domain the function is a disjoint union of open intervals, we say that <m>f</m> is continuous on its domain if it is continuous on each interval.
  </p>
</aside>

sean-fitzpatrick:
I added the following, next to the theorem stating which functions are continuous on their domains:
```
  <aside>
    <p>
      We have defined what it means for a function to be continuous on an interval,
      but many functions, such as <m>f(x)=\tan(x)</m>,
      have domains that are the union of more than one interval.
    </p>

    <p>
      If the domain of a function is a union of intervals, saying that a function is continuous on its domain
      means that the function is continuous on each of those intervals.
      But be careful to note that the converse is not true.
      As we learned in <xref ref="ex_contint2"/>,
      a function can be continuous on a collection of intervals, but not on their union.
    </p>
  </aside>
```

sean-fitzpatrick:
I didn't include "disjoint" (unintentionally), or "open" (intentionally, because we can construct examples -- like square roots of polynomials -- where the domain is a disjoint union of closed intervals)

APEXCalculus:
Yes - I forgot that step. Thanks.

PR #168, opened 2022-03-15
Clairfy closed interval terminology

sean-fitzpatrick:
Address Issue #165.

### DIFF
--- a/ptx/sec_graph_extreme_values.ptx
+++ b/ptx/sec_graph_extreme_values.ptx
@@ -491,12 +491,12 @@
   </figure>
 
   <p>
-    <xref ref="thm_extreme_val">Theorem</xref>
+    <xref ref="thm_extreme_val"/>
     states that a continuous function on a closed interval will have both an absolute maximum and an absolute minimum.
     Common sense tells us <q>extrema occur either at the endpoints or somewhere in between.</q>
     It is easy to check for extrema at endpoints,
     but there are infinitely many points to check that are <q>in between.</q>
-    <xref ref="thm_criticalpts">Theorem</xref> tells us we need only check at the critical points that are in between the endpoints.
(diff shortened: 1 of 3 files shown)


sean-fitzpatrick:
@APEXCalculus please let me know if these changes are sufficient to address #165. If not, I can make additional changes as needed.

sean-fitzpatrick:
Actually there is a related item brought up by @Alex-Jordan in #156. Should we clarify what is meant by a function being "continuous on its domain" when that domain is not connected?

We want to say something like, "If the domain is a union of intervals, a function is continuous on its domain if it is continuous on each interval." Except...

- The domain of the step function is all real numbers, but we can write this as a union of half-open intervals, on each of which the function is continuous
- A function like `\sqrt(x^2-1)` has domain `(-\infty, -1]\cup [1, \infty)`, which most students would consider to be a union of half-open intervals. (We do not want to start talking about the subspace topology...)

So if we wanted to formalize the definition of "continuous on its domain" we would need something that can distinguish between these two examples without getting into some sort of topological quagmire. (I guess the difference between the two is that the intervals in the first example cannot be separated by open sets, but how do you explain that in Calc 1?)

sean-fitzpatrick:
Sorry, one last question: do we want to introduce the language "closed and bounded" for intervals of the form [a, b]?

I don't think "bounded" shows up in the single-variable material, but it's what we use for the Extreme Value Theorem in several variables.

APEXCalculus:
Should the current pull request be merged, then another submitted to make a change similar to what I suggested? What's the standard way of adjusting a pull request?

sean-fitzpatrick:
This one is my pull request, so the easiest way to adjust is for me to add a new commit, based on your suggestion. I'll do that now :-)

PR #176, opened 2022-05-08
tweak WW randoomization

Alex-Jordan:


Alex-Jordan:
This exercise was allowing for questions like:

Find the derivative of ln(-6x^3).

Not really doable given the limitations of only knowing what is known in this section. This change forces the coefficient to be positive. Also the exponent on x, just because.

Alex-Jordan:
I created this branch on the APEXCalculusPTX repo instead of my own fork. It was easier that way given that I was only working in GH. If this gets merged, we just need to delete this branch too.

PR #177, opened 2022-05-12
Update l'Hospital's rule and updates for schema changes

sean-fitzpatrick:
This pull request takes care of the changes to l"Hospital's rule that were requested in #142 but never followed through on.

There are also a few project-wide changes to address flaws in our markup, or changes to PreTeXt in the past year:

- There were a few exercise groups without introductions, that I think were inserted just to get a two-column layout. I've either added an introduction or removed the exercise group, as seemed appropriate
- URLs now need to have a `@visual` attribute. This has been addressed.
- tabulars can no longer go in a figure; they have to be in a table. As needed, figure/caption has been changed to table/title
- the syntax for specifying the format of a list appears to have changed from `label` to `marker`. I've made this change.

### DIFF
--- a/ptx/appendix_back_reference.ptx
+++ b/ptx/appendix_back_reference.ptx
@@ -585,7 +585,7 @@
       <sidebyside>
         <p>
           <alert>Triangles</alert>
-          <ul label="">
+          <ul marker="">
             <li>
               <p>
                 <m>h=a\sin \theta</m>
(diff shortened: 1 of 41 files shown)


APEXCalculus:
Reviewed L'Hopitals content & it looks good. Other content is likely ok, trusting that the PreTeXt material was right.

PR #181, opened 2022-05-19
Set up APEX to build using PreTeXt CLI

sean-fitzpatrick:
This pull request changes almost every file, but changes are all structural: there are no content changes.

Changes are as follows:

- directory reorganization: external images are now in an 'assets' folder
- removal of obsolete files: this includes some mathematica files that haven't been touched in 3 years, as well as the makefile. I think the CLI renders the makefile obsolete
- creation of project.ptx manifest file
- updates to the publication file to support directory management and the CLI
- updates to the extra XSL
- setup for versioning: all video-related elements now have a `component="video"` attribute, and inclusion of videos can be set in the publication file. The color vs print option is also now controlled by the publication file.
- the project manifest has separate build targets for html, html without video, latex, latex without video, latex for print on demand, and latex for print on demand without video
- an updated README file to explain how to build the book using the CLI

There is no longer any need to make changes to the publication file, or anything else, to build different versions. Just pick the appropriate build target from the manifest.

Other notes:
- there is some preliminary work toward EPUB generation, by adding a cover page, and updating the publication file, but EPUB still isn't really supported
- there is a `youtube.txt` file that I hope will be removed at some point, explaining the edits I currently make to the .tex file to make the QR codes for YouTube videos fit in the margin correctly
- I'll probably make a few more small adjustments, but so far everything passes testing


APEXCalculus (commented):
I think all this sounds good; I'll let someone else who knows PreTeXt better actually approve the pull request.

sean-fitzpatrick:
I am really happy with how much easier the build process is now.
You want html? Run `pretext build html -w -d`.
(Need to update the HTML, but webwork and diagrams are unchanged? Run without the `-w -d`.)
You want it without videos? Run `pretext build html-novid`
You want LaTeX for print on demand? Run `pretext build latex-print -w -d`
and so on.

The remaining hang-up is getting the PDF to look right.
Right now we still need significant hand-editing of the .tex file to place margin figures correctly,
and there are the adjustments that I make to the QR codes for the YouTube videos to ensure they fit nicely in the margin.

sean-fitzpatrick:
@Alex-Jordan let me know if this needs some more work.
I think version 0.6.4 of the CLI removes the need for the `entities.ent` file, so if there was concern about that, it can be removed.

I can also remove the text file with the notes on how to tweak the .tex file to get video QR codes to look right in the margin, although we need to come up with a better long term solution for this.

sean-fitzpatrick:
I should add a summary of file changes:
- deleted old Mathematica files
- deleted makefiles (not needed for CLI)
- moved contents of `external-images` to `assets` folder
- changes to extra XSL and publisher files
- added project manifest
- temporary addition of `entities.ent` file
- included notes on hand-editing currently needed for the `.tex` file
- for most `.ptx` files, changes are restricted to adding the video component attribute
- exception is `sec_taypoly.ptx`: I used to keep a second version for use in my version of the books, where we introduce Taylor polynomials much earlier. Now that we have versioning, I was able to combine everything into a single file, containing two versions.

sean-fitzpatrick:
@Alex-Jordan I think the webwork server specification has moved to the publication file since I opened this pull request. I'll make that change and push if it passes testing. Steven is working on the `entities.ent` issue, but at the moment we still need to include it.
Once this is taken care of, we can start looking at cleaning up the LaTeX images (all the places where there are issues with labels, etc.)

Alex-Jordan:
The `@webwork-problems` attribute should be removed from the publication files. Since `@generated` is in there, it relies on that for location of the representations file.

Alex-Jordan:
There is too much here for me to do an honest review. So I think I'm just going to merge and see how things go!

Some recent work will let us get tikz images back into all the exercises. I'm going to see if I can get that started now. (But it won't work with the CLI until the usual lag catches up.)

Issue #161, opened 2022-02-17
Disk/Washer Exercise Graph Labels


grady:
In Section 7.2 on Volumes using Disk/Washer method. In many problems, the equation of the curve to be revolved are shown inside the plot. The positioning of the function equation label needs adjusting in many cases. In particular _problem 4 and 8 have the power of 2 completely cut off,_ displaying a linear equation for what is obviously a quadratic curve.

Even better than adjusting the position of the equation label in the plot, I think the equation and bounds of the region to be revolved should be given in text. This would be better from an accessibility perspective as well.

https://opentext.uleth.ca/apex-video/sec_disk.html

grady:
This problem also occurs in the exercises for the Shell Method in the next section.

Alex-Jordan:
One could try to fix these by changing how cropping works, so that it would now crop to encompass all label text too. This would have the side effect of making adjacent or nearby images seem misaligned. Visually, the axis windows would differ each time a label extended outside of the axis window.

One could try to fix these by visiting each image and adjusting the positioning of these labels until they are within the axis window. That would be tedious and prone to error. It could still cause problems in the future if there is something like a font size change, which might move something to once again be cropped or otherwise mess with carefully laid out positioning.

One could try to fix these by using an `\addlegendentry` to label these curves instead of putting node labels directly on the image somewhere. This would also be tedious. More tedious than above, because you would probably want to do this for all images for the sake of consistent styling. So not just the currently misbehaving images. But I think this is what I lean toward out of the options I have considered. 

Are there other ideas? Or a different preference among these three?

grady:
I think the function definition should be presented in the text, not in the images. Something along the lines of 

> Find the volume enclosed when the curve $f(x)=3-x^2$ is revolved about the $x$-axis.

with the graph attached as a visual aid.

sean-fitzpatrick:
OK, back at work. But I guess I will be focused on things like "teaching my students" and "salvaging the semester", so unlikely to turn my attention to this any time soon.

Seems like this is a bigger project that should be systematically addressed:

1. Try Alex's suggestion of adding `\addlegendentry` to the styling for labels throughout. (Can this be done in global styling?) Presumably we should test to make sure this solves the problem before proceeding ;-)
2. Identify all the places in the book where there is information in an image that should be repeated in the text of the exercise.

[...]

PR #192, opened 2022-08-15
Adjust image labels and add text below images to describe region

sean-fitzpatrick:
Addresses the issue noticed by @grady in #161.

I've added text below each image to describe the region, and I've also adjusted the node shifts for the labels so that they fit in the image, and bit better with the curves than before.

### DIFF
--- a/ptx/sec_disk.ptx
+++ b/ptx/sec_disk.ptx
@@ -1801,8 +1801,8 @@
 
         <introduction>
           <p>
-            In the following exercises, a region of the Cartesian plane is shaded.
-            Use the Disk/Washer Method to find the volume of the solid of revolution formed by revolving the region about the <m>x</m>-axis.
+            Use the Disk/Washer Method to find the volume of the solid of revolution
+            formed by revolving the given region about the <m>x</m>-axis.
           </p>
         </introduction>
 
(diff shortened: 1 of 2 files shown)


APEXCalculus:
@sean-fitzpatrick @Alex-Jordan This looks good to me, but it delves into things I'm not so good at. So I'll let someone else merge this.

Alex-Jordan:
It strikes me as not right to have these image descriptions be a `p` that follows the `image`. It's not "right" and also it could (will?) lead to ugly indentation. These are short enough, can they do into a `description` element? Presently that becomes alt text in HTML. There's been some talk to change `description` to some other mechanism (where it can keep markup like math) and have a new `short-description` that could become alt text.

I don't want to just pop in to be a naysayer. I'm on an extended road trip and I can't help with anything substantial. If you need to do something now, I am not sure the current available alternatives would be better.

sean-fitzpatrick:
I am fine with putting things in a description, although as you say, currently that won't allow us to use math markup.

The suggestion to put the description in a paragraph might have come from @grady ?

In any case, I've made sure that none of the labels are cut off now, but there needs to be something there to ensure these exercises can be done by someone with visual impairment.

The advantage of a paragraph (at present) is that we can use math, and let MathJax handle accessibility.

sean-fitzpatrick:
I've made a quick adjustment to modify the paragraph slightly, and put it above the image. This seems a little better to me?

I forgot to remove a few instances of "In the following exercises". Let me get to that

sean-fitzpatrick:
Does anyone have an opinion on whether this is acceptable, at least as a temporary fix?

Throughout Chapter 7, there were exercises where you could only determine the function by looking at the image.
I've fixed the images so that labels are not cut off, so this, at least, is now possible.
But I've also added text preceding the image in each exercise to describe the function/region.

Maybe this should become alt text later, but I think it's OK. I hope to have PDFs up this weekend for inspection.

PR #195, opened 2022-08-15
Update structure and wording for numerical integration exercises

sean-fitzpatrick:
This is intended to fix the issues identified in #152.

It's possible further work is needed, so suggestions are welcome.

### DIFF
--- a/ptx/sec_numerical_integration.ptx
+++ b/ptx/sec_numerical_integration.ptx
@@ -2095,437 +2095,505 @@
 
         <introduction>
           <p>
-            In the following exercises, a definite integral is given.
-          </p>
-
-          <p>
-            <ol>
-              <li>
-                <p>
-                  Approximate the definite integral with the Trapezoidal Rule and <m>n=4</m>.
-                </p>
-              </li>
-
-              <li>
-                <p>
-                  Approximate the definite integral with Simpson's Rule and <m>n=4</m>.
-                </p>
-              </li>
-
-              <li>
-                <p>
-                  Find the exact value of the integral.
-                </p>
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
At a glance, I don't see anything wrong with the webwork/task structure. And if you are getting good output in the HTML with active versions of the problem, and getting good static PDF, and also running validation checks, it's hard to imagine it's not good.

In the exercisegroup/introduction~s here, there is wording like "In the following exercises...". (You did not add that; it was already that way). But the wording should consider that when an individual WW exercise is presented (say to a student actually using WeBWorK) the introduction and the individual exercise will be presented together. So then the plural case will seem wrong to a reader. On the other hand, I think it's OK to have the singular case when reading the book itself with all the exercises there. The "In the following exercises..." can be implied. It can even be automated to have a heading like "Exercises 11–20." 

So if you feel like it, these things could be worded like "Approximate the definite integral with the Trapezoidal Rule and Simpson's Rule with <m>n=4</m>. Then find the exact value."

sean-fitzpatrick:
I must have posted on an issue for this: the only problem is when you activate things as a webwork problem -- none of the `<var/>`s become answer blanks.

I think the markup is OK, so maybe this is a processing problem?

Alex-Jordan:
I narrowed it down to a commit from June 29. I may not be able to look more closely for a while. I posted to the forum.

sean-fitzpatrick:
I saw the post, thanks. Not urgent. We often skip that section anyway!
(We really shouldn't though -- it's more relevant than any of the techniques in the chapter that follows.)

PR #197, opened 2022-08-16
Move videos in solutions to separate solution

sean-fitzpatrick:
Here is a pull request to put video solutions into a separate solution.
Don't merge this yet, because this might mess up our PDF, in which case it's not worth it for a minor improvement to the HTML.


APEXCalculus:
Won't merge per the request.

sean-fitzpatrick:
I've now updated the XSL to handle having videos in a separate solution.
The video solution gets moved to the margin, with title "**Solution 1** (video solution)" and then the text solution runes inline as usual, with the title "**Solution 2**". (The numbering is added by standard PreTeXt. If we don't like it, there's probably some way to override it.)

I've also figured out how to automatically handle all of the adjustments I used to have to do by hand to make sure things fit nicely in the margin: horizontal shift of the QR code, and changes to the URL caption.

The XSL is maybe a bit of a hack but I'm happy with how it turned out!

sean-fitzpatrick:
This should be my last adjustment to the LaTeX style sheets. Previously we had no need/desire to move `table` to the margin, because all of our `tabular` were in Figures. But Rob changed the schema to ban tabular inside of figure, so now there are a bunch of tables that need to go in the margin.

I've also corrected an earlier error, where if there was a figure containing a video inside of a solution, other than the main solution video, it was being caught in the video solution filter.

This should now be fine to merge, as long as everyone is happy with having a separate `<proof>` for videos.
Of course, in the non-video version, nothing is changed by this pull request, except for the treatment of tables.

sean-fitzpatrick:
I've resolved the conflict (which was very minor) in the requirements.txt file -- this just keeps track of what version of the CLI should be used to build the book.

PR #199, opened 2022-08-19
update publication files to use separate webwork courses for colour/print

sean-fitzpatrick:
When WeBWorK images are generated, the TikZ settings come from the APEX.pl file rather than the publisher file or docinfo.

So if you are trying to produce a print version, the webwork server will not give you black and white images.

I've set up two different courses on my webwork server; one uses colour and the other does not.
This updates the publication files to select one course or the other, depending on what you want the images to look like.

### DIFF
--- a/publication/publication-novid-print.ptx
+++ b/publication/publication-novid-print.ptx
@@ -8,7 +8,7 @@
         <version include="print taypoly-late"/>
     </source>
 
-    <webwork server="https://webwork-dev.uleth.ca"/>
+    <webwork server="https://webwork-dev.uleth.ca" course="apex-print"/>
 
     <common>
       <tableofcontents level="3"/>
--- a/publication/publication-novid-print.ptx
+++ b/publication/publication-novid-print.ptx
@@ -4,7 +4,7 @@
 
     <!-- Relative path, relative to "main" source file -->
     <source>
-        <directories external="../assets" generated="../generated-assets"/>
+        <directories external="../assets" generated="../generated-assets-print"/>
         <version include="print taypoly-late"/>
     </source>
 
--- a/publication/publication-novid.ptx
+++ b/publication/publication-novid.ptx
@@ -8,7 +8,7 @@
         <version include="color taypoly-late"/>
     </source>
 
-    <webwork server="https://webwork-dev.uleth.ca"/>
+    <webwork server="https://webwork-dev.uleth.ca" course="apex-main"/>
 
     <common>
       <tableofcontents level="3"/>
--- a/publication/publication-print.ptx
+++ b/publication/publication-print.ptx
@@ -8,7 +8,7 @@
         <version include="print video taypoly-late"/>
     </source>
 
-    <webwork server="https://webwork-dev.uleth.ca"/>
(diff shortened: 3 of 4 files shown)


Alex-Jordan:
I think I can see a way to handle this more robustly using only one host course. Will you be at drop-in next Friday?

Anything is fine for APEX right now, as long as it works for you. More generally for multiple `latex-image-preambles` on a project, I think we can pack them all into one course's `PROJECT.pl` file and only use the right one at the right time with the right publisher variables.

sean-fitzpatrick:
I'll try to come to the next drop in. If you have a specific time in mind, I can try to have that free.

I figure it costs me nothing (other than about 2 minutes of my time) to create an extra WeBWorK course for print. (In fact I've also made several other courses for my local versions, so I don't have to worry about cached image files.)

But long term, a better solution would be good.

I like having two generated assets folders though, so I don't accidentally use black and white Asymptote for the colour version, or whatever, and I don't have to rebuild all assets each time.

sean-fitzpatrick:
@Alex-Jordan I can't recall if we ever got around to discussing options for this.
I don't need it for my own builds, since I have separate publication files (not tracked) for most of the versions I produce.
So I could close this, but if someone wants to build print PDF from the main repo, I'm not sure we're set up for it.

Alex-Jordan:
Sorry I haven't been very responsive. I won't be at drop-in today, but next week? We could make a point to discuss it then.

sean-fitzpatrick:
I can try to make it. School here ends at noon on Fridays, and after-school programming doesn't start unti February.

sean-fitzpatrick:
Adding a periodic reminder for @Alex-Jordan that we are supposed to discuss this at some point. I don't think it costs me anything to provide a second course for different APEX builds, but if there's some potential problem with this approach, I am open to other possibilities.

Alex-Jordan:
I think you should go ahead with whatever is working for you. I'm not sure that I was thinking of anything that could be done now, without more development first. And who knows when that could happen at this point.

sean-fitzpatrick:
Since I have another pull request with some more urgent changes to the publication file, I suspect there will be conflicts between this PR and that one. Might be better for me to add some of the changes here to that pull request.

sean-fitzpatrick:
I double-checked on a branch that we can merge this to get the changes here without reverting the other recent changes to the pubfile.
@Alex-Jordan and I briefly discussed this, and it seems that any "cleaner" solution would require writing new code that nobody has time to write.

To recap: this PR sets things up so that the publisher can simultaneously build HTML, colour PDF, and print PDF, without having to rebuild webwork, images, asymptote, etc. every time.

Pros:
- streamlines workflow (less time to build each version of the book)
- eliminates chance of accidentally forgetting to rebuild all the assets (which I have done many times)
- doesn't require any new code

Cons:
- adds a couple of lines to the publisher files
- publisher will now have two generated assets folders instead of one on their local machine

Since storage is cheap, this particular publisher says that the last is not a big deal.

sean-fitzpatrick:
Mergin, since we can't come up with a better way to do this.

PR #204, opened 2022-10-10
fix some webwork code

Alex-Jordan:


sean-fitzpatrick:
Assuming you've successfully built webwork representations, I'm willing to merge as is. I don't think I'll have time this week to test the changes.

Alex-Jordan:
I haven't made WW reps, but I did make the corresponding edits to our pg files directly and they are working. So it would only be if I made some sort of typo with the XML here.

PR #207, opened 2022-11-18
enable search

sean-fitzpatrick:


sean-fitzpatrick:
Realizing now that this will no longer be necessary: search is now enabled by default with the most recent version of the CLI.

PR #211, opened 2022-12-08
Pubfile update

sean-fitzpatrick:
I think this should update our publication, project and style files with regard to the recent work to migrate most remaining string parameters to the publication file.

### DIFF
--- a/project.ptx
+++ b/project.ptx
@@ -11,7 +11,7 @@
       <source>ptx/apex.ptx</source>
       <publication>publication/publication.ptx</publication>
       <output-dir>output/html</output-dir>
-      <stringparam key="webwork.divisional.static" value="no"/>
+      <!-- <stringparam key="webwork.divisional.static" value="no"/> -->
       <stringparam key="html.annotation" value="hypothesis"/>
     </target>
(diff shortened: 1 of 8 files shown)


Alex-Jordan:
If this is working for you, I say go for it and merge it. I'm not in a position where I could test things right now.  I skimmed it and didn't see anything to wonder about.

sean-fitzpatrick:
Assuming there are no mistakes in the new documentation, this should be fine. I haven't done a LaTeX build yet, however. I'll do that this afternoon. (Keeping in mind that this will not include the follow-up work to make all the changes, like margin figure placement, that we're not able to encode in our source somehow.)

sean-fitzpatrick:
I've added a line to the publication files for HTML that enables the "CRC" layout (centred, with round corners).

If you'd like to see what this looks like, I've just put a [copy up on the web](https://opentext.uleth.ca/apex-crc).
It's possible not everyone will prefer this layout, but having it in the publication file seems better than not: anyone who doesn't like it can just comment out that line.

sean-fitzpatrick:
I'm going to merge this, which will make CRC the default

PR #218, opened 2023-01-16
Publication file updates

sean-fitzpatrick:
Two updates to the publication file:

1. Change the depth for the table of contents from 3 to 2, so that subsections are not included for the 'official' APEX builds. (The versions I build for use here have parts, so 3 is the correct depth for them.)
2. Update the way the front cover is specified in the publication file, due to a recent change pushed by Rob.

While I was at it, I figured we might as well include the cover for PDF, so I added a line for the cover in the `<latex>` section as well. (The cover itself is already committed, but I had to move it into the "assets" folder.)

If nobody likes the cover image, we can (a) update it, or (b) omit it. The latter I can do by pushing one more commit to this pull request to omit it from the PDF.

We're not building epub yet anyway so this is a future problem.


sean-fitzpatrick:
Since this is a month old, I'm giving it a bump.

Probably I should have put the change to the ToC depth on a separate pull request. 
Since it's too late for that, does anyone have an opinion on the inclusion (or not) of the cover image?
This would be added as the first page in the PDF build.

Here's what it looks like:
cover

APEXCalculus:
I'm not a big fan of the cover image. I know this is a personal preference
thing, so perhaps I'm just pitting one opinion against another.

Every time I tried to put an image on the cover that I liked, I failed. It
never looked as good to me as I hoped. That's a big reason the current
print versions have such bland covers - I felt it was better to keep it
simple.

And you are right - I missed this part of the previous pull request and
only thought about the TOC.

Greg

sean-fitzpatrick:
The cover is admittedly not best work ;-)

I originally made it about a year and a half ago, to test out whether building epub would work.
(Epub requires a cover image.)

I'll push an update to comment out inclusion for PDF, and we can put it back if we come up with a good cover some day.
It needs to stay put for epub, but since we aren't trying to build epub yet, we have time to come up with something better.

sean-fitzpatrick:
I have the changes to the publisher that are needed to ensure nothing breaks on another branch, so I am closing this PR.
At some future date I will try to come up with some better cover ideas.

PR #223, opened 2023-05-01
Add labels to videos for Runestone

sean-fitzpatrick:
YouTube videos currently don't work in the Runestone build because labels are missing.
This adds labels to all YouTube videos.


Alex-Jordan:
Hope that works.

You could ask Brad for author access to APEX on Runestone Academy. Then there is a page you can go to to rebuild the book from wherever the default GitHub branch is at.

PR #226, opened 2023-05-17
May2023 edits (phase 1)

sean-fitzpatrick:
Creating a PR now before I have too many changes on this branch.

What's in this pull request:

1. Updates to the publisher file:
     - Change TOC level from 3 to 2 (show sections but not subsections)
     - Move WeBWorK configuration from string parameter in `project.ptx` to publisher file
     - Change syntax for epub cover to match recent changes (even though we don't use this right now)
2. Add document id as requested by David
3. Change two exercise groups with 3 columns to have 2 columns
     - Includes adding an extra exercise in `sec_deriv_inverse_function.ptx` to have an even number
     - Also includes removing a duplicate exercise in  `sec_limit_onesided.ptx`, for the same reason
4. Change most tables back to figures. This includes updating xrefs so that 'Figure' or 'Table' is not hard-coded, in case we change our minds...
5. While fixing the table thing, I noticed that two examples in `sec_graph_extreme_values.ptx` have references in the statement to figures that do not appear until the solution. I decided it made the most sense to just remove the references, and leave the figure in the solution.


sean-fitzpatrick:
I realized that the branch I was using for this didn't have the last PR that added labels to videos.

I merged that change, then also realized that I probably should have rebased rather than merged. (But I think the difference is mainly in what the commit history looks like -- the resulting changes to the text end up the same.)

sean-fitzpatrick:
I am probably going to split this up into a few smaller pull requests... stand by

sean-fitzpatrick:
Closing this to replace it with some smaller pull requests that hopefully can be more quickly approved.

PR #229, opened 2023-05-26
Update some WeBWorK code

sean-fitzpatrick:
Biggest change is updating applied problems in Chapter 3 to use TikZ images.
Previously these problems used pg graphs, or did not use WeBWorK.

Building revealed a deprecation: a macro was being called that didn't need to be, so I removed references to that macro.

I also updated a question using checkboxes to use recent code that's properly supported.


sean-fitzpatrick:
Reviewed during dropin

PR #230, opened 2023-05-26
Housekeeping updates

sean-fitzpatrick:
Pulling some small changes out of larger pull request #226 (which is now closed).

- Update TOC level in publisher file. This has been in a few pull requests that I've closed because they got ignored for months and went stale. Official APEX doesn't use parts, so setting the TOC level to 3 exposes subsections in the side panel, and it looks bad. (My versions have parts, but I have my own pubfile that is not tracked by git.)
- Get rid of a string parameter for webwork that is now in the publisher file
- Update `requirements.txt` to the current ptx version
- add a document id, as requested by David
- make sure the publisher file handles epub correctly

Note that for the last item, we are still not ready to build epub; this just ensures that we don't get warnings/errors while trying to build other targets.


sean-fitzpatrick:
Alex says it's OK to merge ;-)

PR #234, opened 2023-06-01
Update extra XSL for LaTeX/PDF

sean-fitzpatrick:
Changes to PreTeXt over the last year rendered our extra XSL useless.

There were minor changes to the figure and image templates, and major changes to how videos are handled.
This updates our XSL to conform to those changes, noting that further updates may be needed, as handling of videos via the PreTeXt assembly process is still in development.

Notes:

- The video template is changed completely. I hijacked the video template in `pretext-assembly.xsl`, but had to write the output to a new element (I called it `margin-video` because there are several passes with assembly, and by the time our figure and solution templates went looking for videos, they were no longer there
- As a result, `figure` and `solution` are now conditioned on `margin-video` rather than `video`
- Since we have (for now) moved tabular back into figure, I've commented out the XSL that moves tables into the margin. Any remaining tables are large enough to be considered proper tables that should not go in the margin
- @APEXCalculus wanted there to be a colour PDF where the page orientation alternates for left- and right-hand pages. I've added an extra style sheet to accommodate this option, as well as new publisher files and `project.ptx` entries.

This means we are now set up to build three types of PDF:
- colour, for use on a laptop
- colour, for printing
- greyscale, for printing

I've run a few tests and the new XSL seems to be working as desired. In fact, there are some improvements to things like the YouTube display URLs when videos are included.

Unfortunately there's still nothing I can do to improve placement of objects in the margin. That still needs to be manually adjusted.

### DIFF
--- a/project.ptx
+++ b/project.ptx
@@ -40,6 +40,20 @@
       <output-dir>output/latex-print</output-dir>
       <xsl>style/apex-latex-print-style.xsl</xsl>
     </target>
+    <target name="latex-color-print">
+      <format>latex</format>
+      <source>ptx/apex.ptx</source>
+      <publication>publication/publication-color-print.ptx</publication>
+      <output-dir>output/latex-color-print</output-dir>
+      <xsl>style/apex-latex-print-color.xsl</xsl>
+    </target>
+    <target name="latex-color-print-novideo">
+      <format>latex</format>
+      <source>ptx/apex.ptx</source>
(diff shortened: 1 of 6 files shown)


sean-fitzpatrick:
Is anyone else interested in attempting a LaTeX build to test out these changes? It is working for me.

I think the main PreTeXt code for videos is still a moving target, but for the moment, this gets the job done.

PR #235, opened 2023-06-13
Clarify location of LaTeX style files

kcrisman:
The readme makes it sound as if there are actual tex files in the source, which there are not.


kcrisman:
If this is unneeded or wrong, just let me know.  Obviously incredibly low priority, but could be useful for clarification.

kcrisman:
Sorry about the extra PR, apparently GH didn't like me renaming the branch.

PR #236, opened 2023-06-13
Clarify location of LaTeX style files

kcrisman:
The readme makes it sound as if there are actual tex files in the source, which there are not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,8 @@ The following variations are also available:
 **Do not build PDF directly.** APEX Calculus places figures in the margins in PDF.
 Unfortunately, we do not currently have a mechanism for setting the vertical placement
 of margin figures in the PreTeXt source, so these have to be adjusted by hand.
-Look in the LaTeX source for occurrences of `\listmarginbox`, `\parmarginbox` and `\tcbmarginbox`.
+Look in the LaTeX XML stylesheet files (in the style directory)
+for occurrences of `\listmarginbox`, `\parmarginbox` and `\tcbmarginbox`.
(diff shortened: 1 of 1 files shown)


kcrisman:
If this is unneeded or wrong, just let me know. Obviously incredibly low priority, but could be useful for clarification.

sean-fitzpatrick:
This could be more clear, but it does indeed refer to the LaTeX source.

APEX puts figures and asides in the margin. I have not been able to find a solution for margin placement in LaTeX that's intelligent enough to automatically place objects in a way that avoids overlaps or running off the page.

So once you've built the .tex file from PreTeXt source, you have to build the PDF, and then visually inspect to see what needs fixing. Every instance of a margin object in the .tex source has a positional argument at the end. The default is `{0cm}`  but this needs to be changed to move each object either up or down as needed.

If there's a way to position things in the margin that avoids this, I would very much welcome that suggestion!

sean-fitzpatrick:
@kcrisman are you using APEX in your classes at all? We still need a couple of people for Vilma's project proposal.

kcrisman:
> are you using APEX in your classes at all? We still need a couple of people for Vilma's project proposal.

I won't be able to myself (very inappropriate for the course I teach), but I might be able to convince some colleagues to do so, especially by the time frame mentioned.  (I just saw all that discussion, was off line this morning.)  No guarantees, though I think it would be a good move, personally.

sean-fitzpatrick:
If you can, please send an email around ASAP -- I think Vilma needs to submit by the end of  this month. Thanks!

PR #244, opened 2023-06-16
Add labels to exercises

sean-fitzpatrick:
Adds `@label` elements to all exercises, which will allow the book to work on Runestone.

Kind of a big PR, since there are something like 3500 exercises across 90+ section files.

### DIFF
--- a/ptx/sec_ABC.ptx
+++ b/ptx/sec_ABC.ptx
@@ -2013,9 +2013,9 @@
                 </image>
               <!-- figures/fig_07_01_ex_29.tex END -->
 
-              <p>
+              <!-- <p>
                 <var name="$area_ft" width="20"/>
-              </p>
+              </p> -->
    <!-- 262,800 ft^2 -->
             </statement>
             <answer>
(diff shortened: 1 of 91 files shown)


sean-fitzpatrick:
@Alex-Jordan I think it would be a quick regex find and replace to copy each of these labels onto the `<webwork>` element as an `xml:id`. That would give us an `xml:id` on every `webwork` (something we've talked about in the past), which should help ensure that exercises remain stable.

In the currently published HTML, I've noticed that somewhere in the book an exercise must have been added, and that's pushed all the exercise groups to be off by one: e.g. the first exercise in exercise-group 2 should be the last exercise in exercise-group 1. I think having xml:ids would help avoid this.

sean-fitzpatrick:
Can confirm that adding ids to webwork elements was easy. One find+replace with regex plus a couple of strays.

sean-fitzpatrick:
I've gone through the build process and everything seems to be OK with these changes. If nobody has any objections, I'd like to merge so we can get a functioning version of APEX on Runestone.

I know it's a big PR in terms of reviewing.... 3500 exercises to add labels to across 91 files means a lot of changes.

PR #247, opened 2023-06-19
Errata fix3

APEXCalculus:
Another smattering of errata, many from the Axel email.


sean-fitzpatrick:
You may want to reconsider the change from `bluepen` to `blue` in the Asymptote code in the cross product section.

The reason this is there is because we want to be able to produce both colour and greyscale versions.

There is an `asymptote-preamble` in the `docinfo` file. In the publisher file, we can choose to include either the `color` compoent, in which case `bluepen=blue`, or the `print` component, in which case `bluepen=rgb(.1,.1,.1)`.

If the colour is hard coded as `blue`, it will remain blue in the black and white print version.

APEXCalculus:
The 'bluepen' to 'blue' conversion wasn't intentional - nice catch. The graphic shows a parallelpiped with three of the edges labeled, but one of the labels was at "the back" and was hard to read. My true intent was to reposition certain vectors and their labels, but got careless with some cut/paste operations.

sean-fitzpatrick:
This now looks good to me.

PR #251, opened 2023-06-27
Minor typo and wording changes.

APEXCalculus:


sean-fitzpatrick:
This is giving a merge conflict because you had a recent pull request that changed `(1/4)^n` to `(1/4)^(n+1)`, and that PR was not merged into your branch before correcting this to `(1/4)^{n+1}`.

It should be easy enough to resolve the conflict an merge if you have no further changes planned on this branch.

If you have other changes you want to push for this PR it might be a good idea to first update your `main` branch and then rebase off of `main`? (With the disclaimer that I've never quite figured out the right way to do a `git rebase`.)

sean-fitzpatrick:
It looks ok to me. Maybe @Alex-Jordan or @davidfarmer can check to make sure this isn't going to cause problems.

Some of the "changes" being shown have already been made. I think the end result is the same and the only issue is that the git history is a bit messier than it otherwise might have been

ptx/sec_vector_intro.ptx
```diff
@@ -420,7 +420,7 @@
 
                 <li>
                   <p>
-                    The scalar product of <m>c</m> and <m>\vec v</m> is the vector
+                    The scalar multiplication of <m>c</m> and <m>\vec v</m> is the vector
```
Alex-Jordan, line 423: Just a comment as long as I am reading; feel free to ignore.

This phrasing now sounds awkward. How about just `The product of <m>c</m> and <m>\vec v</m>...`? If you change, also a few lines down.

ptx/preface.ptx
```diff
@@ -116,9 +116,9 @@
 	</paragraphs>
 
 	<paragraphs xml:id="pars-apex">
-		<title><m>\apex</m> <ndash/> Affordable Print and Electronic teXts</title>
+		<title>APEX <ndash/> Affordable Print and Electronic teXts</title>
 		<p>
-			<m>\apex</m> is a consortium of authors  who collaborate to produce high quality, low cost textbooks.
+			<alert>APEX</alert> is a consortium of authors  who collaborate to produce high quality, low cost textbooks.
```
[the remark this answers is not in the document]
sean-fitzpatrick: This one is my fault! I pushed a quick change to get rid of the broken code, and it got included in Greg's pull request when he rebased.

I didn't know about the pubtitle tag. I chose alert with the thought that we want to draw attention to the title.

We briefly discussed implementing a tag for APEX but I think my to-do list is too long to prioritize it.

Alex-Jordan:
I left two comments that anyone should ignore if thy are so inclined. I didn't verify the mathematical edits here but otherwise I saw no issues.

ptx/sec_vector_intro.ptx
sean-fitzpatrick: I think the intent from @APEXCalculus is to introduce the term "scalar multiplication" here. Maybe there's a better way to do it? Something like, "the product of the vector v with the scalar c is known as scalar multiplication"?

sean-fitzpatrick:
Some of these changes were already merged, actually, and appear here because Greg had to rebase to resolve a merge conflict.

But that doesn't mean we can't still adjust!

@APEXCalculus do you want to try changing `<alert>APEX</alert>` to `<pubtitle>APEX</pubtitle>`?

APEXCalculus:
As far as APEX typesetting goes, I'm not particularly concerned. I agree
that 'alert' is tag abuse. I don't know the difference between 'alert' and
'pubtitle'; I'm sure either is fine. To avoid tag abuse, I lean toward
pubtitle.

ptx/sec_vector_intro.ptx
APEXCalculus: I've given new wording in PR #253 (it took me 2 commits to get it right). It says "The multiplication of a scalar <m>c</m> and a vector <m>\vec v</m> is the vector ... ". The sentence below that definition uses the term 'scalar multiplication', which is implied by the definition.

sean-fitzpatrick:
I think everything here is fine, pending changes in #253 . I'll merge this, and then see if that one will go through as well.

Issue #241, opened 2023-06-15
Line integral language


sean-fitzpatrick:
In the introduction to line integrals section, there are a number of places (and in particular, [in the exercises](https://opentext.uleth.ca/apex-calculus/sec_line_int_intro.html#exercisegroup-361) where there is language like "the surface f(x,y)".

I would like to suggest that this should be either "the function f(x,y)", or in some places "the surface z=f(x,y)".

sean-fitzpatrick:
@APEXCalculus are you all right with changing this? If so, which of the two options suggested above would you prefer?

sean-fitzpatrick:
This was taken care of with #252

PR #252, opened 2023-06-28
Update language: function vs surface

sean-fitzpatrick:
This should address the language issues identified in #241, but across the whole book.
(I basically searched the whole book for "surface" and patrolled for cases where a function was referred to as a surface.)

@APEXCalculus please review to see if the changes seem reasonable, and let me know if there are any issues or mistakes.

In a few places the phrase "graph of" was being used as synonymous with "sketch of" or "plot of". I changed a couple of these but probably missed others. (Maybe this is picky, but to me, "graph of f" refers to a set of points, not the illustration of that set.)

### DIFF
--- a/ptx/sec_directional_derivative.ptx
+++ b/ptx/sec_directional_derivative.ptx
@@ -757,7 +757,7 @@
       <title>The flow of water downhill</title>
       <statement>
         <p>
-          Consider the surface given by <m>f(x,y)= 20-x^2-2y^2</m>.
+          Consider the surface given by the graph of <m>f(x,y)= 20-x^2-2y^2</m>.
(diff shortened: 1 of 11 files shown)


APEXCalculus:
All looks good except for one change. Is it best to merge then fix that one line, or fix with a follow-on commit?

In ptx/sec_surface_area.ptx, line 821, the new line reads "Let f(x,y) and g(x,y) = 2f(x,y).", but that isn't a sentence. 

How about: 
"Let f(x,y) be a function defined over a region R, and let g(x,y) = 2f(x,y). Why is the surface area of z=g(x,y) over R not twice the surface area of z=f(x,y) over R?"

sean-fitzpatrick:
Thanks for catching that! I'll push a correction after breakfast.

PR #255, opened 2023-07-04
Modify webwork ids to be different from exercise labels

sean-fitzpatrick:
In an earlier commit, I used the labels added to exercises to create xml:ids for webwork elements. This should have been fine, but somehow it led to two elements in HTML getting the same ID. (I didn't think `@label` was used for HTML IDs.)

As a result, the Activate button for WeBWorK problems was not working correctly. This modifies the webwork xml:ids to be different from the labels, which fixes the issue with the activate button not working.

### DIFF
--- a/ptx/sec_ABC.ptx
+++ b/ptx/sec_ABC.ptx
@@ -702,7 +702,7 @@
     <subexercises xml:id="TaC-area-between">
       <title>Terms and Concepts</title>
       <exercise label="ex-area-between-TaC-1">
-        <webwork xml:id="ex-area-between-TaC-1">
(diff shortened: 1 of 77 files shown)


sean-fitzpatrick:
Back from holiday, and I guess this is now a stale PR. It looks like a lot of changes, but the summary of the commit is:
- add a `webwork-` preface to every `xml:id` on a `<webwork>` element.
- put labels on one exercise with tasks that got overlooked before.

If nobody has any objections, I'll merge this so I can move on to other things.

APEXCalculus:
I had left this one alone as it used the word "webwork". But it seems a straightforward commit, so I'll merge.

Issue #250, opened 2023-06-27
Limit Composition Exercises


APEXCalculus:
Exercise gp. 1.3 #7-14 has some limit composition issues. As it stands, #11-13 cannot be determined, which isn't the point of these exercises.

The situation can be ameliorated with changing the condition: g(6) = 9 to g(6) = 3 in the instructions.
This change doesn't change any answers that do not involve composition.

Then #11 is answerable with final limit of 3.

We still cannot determine the limit in #12. That's ok.

Then #13 has an answer of 3. 

I think I prefer changing the order of #12 and #13. 

I'd do this myself but these are written in WeBWorK and I'm having difficulty parsing that code.

sean-fitzpatrick:
I'll start work on this.
@Alex-Jordan can you remind me whether or not we now have the infrastructure to include WeBWorK variables in the introduction to an exercise group? If so, we would not need to hard code the values for this problem.

As far as I can tell, we have the ability to pull text from the introduction when building `.pg` files, but not to add variables.

sean-fitzpatrick:
In all of them, it looks like we just have to change `%g = ($q => $p);` to `%g = ($q => $r);`.

Since the code is the same for every problem in this group, I think we could combine them into a single problem with parts. Then we could randomize the initial set of values. For symmetry we could try to come up with 4 more parts for the second set of problems; then in the 2-column layout we would have the two problems side-by-side with the same number of parts.

I'll open a pull request momentarily that fixes the bug, and then I can do some more work if we would like to make this change.

Alex-Jordan:
> include WeBWorK variables in the introduction

You can't do that. If it's variable, then however it comes out in the overall introduction for the exercisegroup in print might clash with how it came out making the individual exercises within that group.

sean-fitzpatrick:
That's what I thought.
@APEXCalculus let me know if you'd prefer to keep this as-is (the numbers will always be the same) or if you'd like to try combining everything into a single exercise with parts.

APEXCalculus:
It seems that leaving it as-is is simplest. This is a pretty low-value problem, in my opinion, as it is so abstract, so leaving it alone seems like a good option.

PR #256, opened 2023-07-19
Fix to address Issue #250

sean-fitzpatrick:
This should fix the problem observed by @APEXCalculus in #250 .
(I still need to rebuild and test.)

I think this is all that is needed, unless we decide that we'd rather combine all of these problems into a single exercise with parts, which would allow for the randomization of the values used to evaluate the limits.


APEXCalculus:
In another thread I suggested leaving this hard coded, as-is, vs. combining into one problem with parts. I re-looked at this and still think its best, but I'm open to push-back. I don't use webwork so the randomization feature doesn't mean much to me. 

If someone else finds it particularly valuable, then yes, these 8 problems could be broken into two sets of 4, then another set of 4 be created for the symmetry (giving 4 sets of 4 problems). Coding one would be the heavy lifting, and the other 3 would follow quickly, I imagine.

Issue #258, opened 2023-07-19
Answer section uses Answer too much and has extra numbering


APEXCalculus:
A partial cut/paste from a comment in another issue:

The word "Answer" is in the answer section, before each answer, seems unnecessary and excessive. Is there a reason to not remove this? Even for questions with multiple parts it seems there are less verbose ways of listing the answers.

I also find the numbering potentially confusing. In Section 6.4, the problems are just numbered 1, 2, etc. In the Answer section, they are numbered 6.4.1, 6.4.2, etc. I don't think that's necessary. Further, in one of the pdf's you sent earlier, in some sections each problem is identified with 4 numbers (such as 6.1.5.1, for Chapter 6, Section 1, SubSection 5, problem 1).

sean-fitzpatrick:
If we only include answers in the back, and not solutions, you no longer see "Answer" printed in the back matter.

The difference in numbering is due to the fact that some sections have subsections, and some do not. I'm not sure there's an easy fix for this. In the original LaTeX version, you got around this by using "decorative" subsections: instead of starting a subsection with `\begin{subsection}`, which would create additional structure, you used text formatting commands to create the appearance of a subsection title without getting the numbering/structure.

PreTeXt is seemingly not as permissive as LaTeX, and insists that a subsection should be structured as such :-)

I'll see if there is some way to tinker with formatting (@Alex-Jordan any ideas?) but it might be the case that we need to structure sections without subsections somehow. (This is where @davidfarmer will chime in to say this is a bad idea.)

sean-fitzpatrick:
I'm going to first try to set the numbering level so we stop numbering at the subsection level (see [the guide] (https://pretextbook.org/doc/guide/html/publication-file-numbering.html)).

If that doesn't work, one option would be to replace each existing `subsection` with `paragraphs`. That might be closer in style to the original, although I think most of what is currently marked up as a subsection really is a subsection...

APEXCalculus:
I'd be in favor of not numbering subsections, though leaving them in the
PreTeXt is fine. I fully acknowledge that I did a hack in the original with
just using some \textbf{\large ...} type structure that I had to look up
again each time I used it - which I don't think was terribly often.

In my opinion, subsections should really follow outline-structure rules,
such as "if there is an A, there should also be a B." I certainly didn't
follow that structure; I'm sure there are sections with only 1 subsection.
It also seems that if some content is in a subsection, then all content
[...]

PR #262, opened 2023-07-27
Adjust numbering of answers in the back

sean-fitzpatrick:
I added some extra XSL to make sure that answers in the back do not get subsection numbering.


sean-fitzpatrick:
This should address the concerns from @APEXCalculus in #258

Issue #261, opened 2023-07-24
Incorrect cropping on graph


sean-fitzpatrick:
See the graph in [this exercise](http://localhost:8000/sec_limits_infty.html#ex-limit-infinity-use-graph-2). The TikZ for this has the same bounds as other exercises in this set, but for some reason, cropping is leaving a bunch of whitespace at the bottom.

I can't spot anything in the TikZ code that explains this.

Here's the code:

```
<latex-image>
  \begin{tikzpicture}
    \begin{axis}[
      xmin = $xmin,
      xmax = $xmax,
      ymin = -50,
      ymax = 50,
    ]
      \addplot[firstcurvestyle, -{Kite}, domain=$xmin:$x[0], samples=100] {1/((x-$a[0])*(x-$a[1])^(2))};
      \addplot[firstcurvestyle, {Kite}-{Kite}, domain=$x[1]:$x[2], samples=100] {1/((x-$a[0])*(x-$a[1])^(2))};
      \addplot[firstcurvestyle, {Kite}-, domain=$x[3]:$xmax, samples=100] {1/((x-$a[0])*(x-$a[1])^(2))};
      \addplot[asymptote,-] coordinates {($a[0],-50) ($a[0],50)};
      \addplot[asymptote,-] coordinates {($a[1],-50) ($a[1],50)};
    \end{axis}
  \end{tikzpicture}
</latex-image>
```

APEXCalculus:
The included link starts with 'localhost' so I can't see what you see.

I substituted values for $xmin, etc., to see what the graphic would create. It looks fine to me. 

I'm assuming that x[0] is a value similar to a[0]-0.01, and x[1] is about a[0]+0.01, etc.? Then the domains make sense. Maybe the error is in there somewhere.

sean-fitzpatrick:
Oh, shoot! Sorry, I usually know better than to do that.
The version [currently online](https://opentext.uleth.ca/apex-calculus/sec_limits_infty.html#exercise-166) seems OK.

The build I just did uses different x values for the two asymptotes, but the graph looks good, except for the fact that it didn't get cropped correctly.

sean-fitzpatrick:
You can see the issue on the [Runestone version](https://runestone.academy/ns/books/published/APEX/sec_limits_infty.html?mode=browsing#ex-limit-infinity-use-graph-2)

APEXCalculus:
[...]

Issue #221, opened 2023-02-01
Comparison for improper integrals with infinite range


sean-fitzpatrick:
Currently the textbook states both direct and limit comparison results for improper integrals, but only for those where x goes to infinity.

For integrals with infinite range (e.g. \int_0^1 \frac{\sin(x)}{\sqrt{x}}\,dx) there is no limit comparison test (that I'm aware of) but direct comparison can still be used.

I think this needs to be mentioned. 
Otherwise, students may make one of two assumptions:
1. there is no comparison test that can be used in this situation
2. both types of comparison test can be used

APEXCalculus:
Yes, we should mention the direct comparison option for these types of
integrals.
I'll add this to my list.
Thanks.


error errata apex calculus

sean-fitzpatrick:
On more careful inspection, there is a paragraph near the end that mentions comparison for integrals with infinte range; it states that this case is more difficult, and omitted from the book.

I added an aside stating that if you really do need to do comparison for an integral of this type, you should stick with direct comparison, as this is less prone to pitfalls. If this is OK, we can close this issue.

Here is my added text:

> If you do need to use comparison for an improper integral with infinite range,
> it is generally wise to stick with direct comparison.
> Direct comparison will continue to work in more or less the way you expect;
> however, limit comparison is much more subtle, and prone to incorrect conclusions.

APEXCalculus:
That sounds good to me.
I looked at this recently when I was addressing my list of typos, etc. I
made a note to myself to "backburner this" as I didn't see a concise way of
presenting it. The comment and your note are enough to say "this is a
thing, but perhaps tricky" and only the most persistent will chase it down.

Thanks.

sean-fitzpatrick:
OK, great! Last semester I noticed that some students were trying to use limit comparison for improper integrals with infinite range, but they were still taking the limit as x goes to infinity! So we had to discuss why that didn't make much sense in that context.

Issue #220, opened 2023-01-30
More exercise errata: differential equation undefined for given intial condition


sean-fitzpatrick:
[Exercise 14 in the section on separable differential equations](https://opentext.uleth.ca/apex-accelerated/sec_Separable.html#exercise-1524) has the differential equation `y' = \frac{x^2}{1-y^2}` with initial condition `y(0)=1`.

This seems like a very bad choice of initial condition, since it results in division by zero.

It seems that we should either change the initial condition, or change the denominator to `1+y^2`.

Alex-Jordan:
This chapter is only in the PTX version, is that right?

If you separate and integrate, even maybe changing the subtraction to addition, you get:

y ± y^3/3 = x^3/3 + C

Are students supposed to solve that for y, with y isolated? If an implicit equation is an acceptable answer, I might suggest that we edit the instructions to let that be known to the student. I'd hate to see students at this level trying to isolate y in the above.

sean-fitzpatrick:
The section on separable equations makes it reasonably clear that implicit solutions are acceptable.

If you ignore the original form of the equation, you can apply the initial condition, so it's possible to "solve" the problem if you sort of blunder forward and don't worry about the initial setup.

Alex-Jordan:
I know I'm nitpicking, but if an implicit solution is OK, then is function notation itself OK? (Let alone the division by 0.) 

In this example, you get:
y - y^3/3 = x^3/3 + 2/3
3y - y^3 = x^3 + 2

Both (0,1) and (0,-2) are on that curve.

sean-fitzpatrick:
I'm just now getting back to this. I think I'm the one who is nitpicking: this curve has a cusp at (0, 1), which is the problem. We don't discuss things like existence and uniqueness of solutions in this book, but I think we should ensure that y' is defined at the point set by our initial condition. Changing the initial contition to y(0)=-2 would address the issue.

sean-fitzpatrick:
I've got this change ready to go on an errata branch I hope to send for pull request by the end of the day.

Alex-Jordan:
(Catching up on what feels like hundreds of APEX, PTX, Runestone, and WeBWorK threads while I was away for a week.)

[...]

Issue #208, opened 2022-11-19
Definition of indefinite integral


sean-fitzpatrick:
We define the indefinite integral to be the set of all antiderivatives of a function.

But [later on](https://opentext.uleth.ca/apex-standard/sec_antider.html#p-4309), we write that `\int f(x) dx = F(x) +C`, where F(x) is an antiderivative and C is a constant.

This isn't consistent. For consistency, we should write `\int f(x) dx = \{F(x)+C | C\in \mathbb{R}\}` but of course nobody does this.

What's the right way to fix this? Maybe we don't definite the indefinite integral as a set, but rather as a "general antiderivative"?
Stewart is no better on this; he says the indefinite integral represents a function, or a family of functions. (Either one, depending on how you feel, I guess?)

Another option is to state that the indefinite integral is not really an object, but a notation:

`\int f(x) dx = F(x)` is short-form for saying, "an antiderivative of f(x) is F(x)," or "the derivative of F(x) is f(x)".

Alex-Jordan:
I think I would want to tell students that in this context, `F(x) + C` is a special shorthand notation for a set. Can we do that?

sean-fitzpatrick:
I think we can do this, but I guess the main question is, how to say something that is consistent, but still makes sense for students?

I guess when we get to solving IVPs, we have an undetermined constant, but this time, it's *an* antiderivative, rather than an indefinite integral.

APEXCalculus:
I think we can "be consistent" if we acknowledge that we are abusing
notation.

Building on Alex's idea, we could write
"Using Definition 5.1.2, we will write that \int f(x) dx = F(x) + C, where
we slightly abuse notation by letting ``F(x) + C'' represent a set."

I'd also like to redo the graphic decomposing the integral notation. I
think it would be wise to add a multiplication symbol between f(x) and dx.
Students tend to think that the integral symbol and the dx act as bookends.

sean-fitzpatrick:
I've made a few changes that I hope to make available soon.

First, I want to note that I am not responsible for the comment in the source that says: "Todo: fix this weird image."
But I did try to fix the image a bit:
- apparently the syntax `right of=node` is deprecated, and should be replaced by `right=of node`. Why do `of` and `=` need to trade places? I don't know. But I swapped them.
- I added a node distance option to tighten it up a bit
- I added an extra node with a `\cdot` to emphasize the multiplication

[...]

PR #263, opened 2023-08-02
Assorted errata

sean-fitzpatrick:
Contains several miscellaneous commits to address several open issues (and fix one previously unnoticed typo).

- graph bounds from #261 have been adjusted
- language discussed in #208 has been added
- an aside warning about our non-standard definition has been added as per #205 
- some of the content proposed by Grady in #150 has been added
- a caution about using limit comparison for improper integrals with improper range has been added (#221)
- the bad initial condition mentioned in #220 has been fixed

### DIFF
--- a/ptx/sec_Separable.ptx
+++ b/ptx/sec_Separable.ptx
@@ -393,7 +393,7 @@
           <p>
             In the following exercises,
             decide whether the differential equation is separable or not separable.
-            If the equation is separable. write it in separated form.
+            If the equation is separable, write it in separated form.
           </p>
         </introduction>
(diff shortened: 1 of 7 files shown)


sean-fitzpatrick:
The remaining issues I'd like to tackle are #200 and #219.
Both have to do with arc-length integrals where the derivative is undefined at the endpoints.

If we can make decisions on how to proceed on those issues, I'll add them to this PR.

APEXCalculus:
I'll merge; I added a comment last night about shortening the incr/decr aside. It isn't a deal breaker, and it wouldn't be hard to change to that with another PR if desired.

PR #264, opened 2023-08-04
In Which the Book of APEX Calculus is Hereby Freed of All WeBWorK-Generated Graphs

sean-fitzpatrick:
Two main items in this pull request:

1. I tracked down all remaining occurrences of `<image pg-name="$gr"/>` (the webwork-generated graphs) and replaced them with TikZ graphs.
2. I found (I think) all places where the WeBWorK code had been commented out because we wanted to keep a TikZ graph in place, made the graph webwork-compatible, and uncommented the webwork code.

Fortunately, not too many of the graphs were in problems with randomization, so my job was a lot easier.
It was also fortunate that I still have Greg's original LaTeX on my computer, so I was able to track down the original TikZ and update it for use with PreTeXt and WeBWorK.

I feel a little bit bad about deleting hundreds of lines of WeBWorK code from @Alex-Jordan to make the old graphs, but I daresay the new ones are nicer. ;-)

Everything compiles, and I've adjusted a bunch that initially didn't look quite right, or had y values that ran off the bottom of the graph. (For some reason, even though you can't see the graph below the value of ymin, the cropping utility is leaving whitespace down to whatever the absolute minimum on the graph is.)

Note that I merged in the changes from the `ol-to-tasks` work before making the pull request, because otherwise there would be quite a few merge conflicts. I've resolved those for us ahead of time.

### DIFF
--- a/ptx/sec_ABC.ptx
+++ b/ptx/sec_ABC.ptx
@@ -774,7 +774,6 @@
     </subexercises>
     <subexercises xml:id="problems-area-between">
       <title>Problems</title>
-    <!--TODO: the following problems need pg-graphs or tikz support in webwork-->
       <exercisegroup cols="2" xml:id="exset-area-between-graph">
 
         <introduction>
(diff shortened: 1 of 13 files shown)


sean-fitzpatrick:
Merged with no conversation. I guess @APEXCalculus is happy to see these go as well :-)

Issue #179, opened 2022-05-16
what is (not) a power series?


Alex-Jordan:
Is the term "radius of convergence" defined only for power series?

If so, then a nitpick is that:
Theorem 9.6.7: The Radius of Convergence of a Series and Absolute Convergence 
https://opentext.uleth.ca/apex-video/sec_power_series.html#thm_abs_power
applies that term to a series that is not a power series.

Or should I interpret
sum a_n |x - c|^n
as the power series
sum a_n y^n
being evaluated at y = |x - c|?

Is item 4 here:
Example 9.6.8: Determining the radius and interval of convergence 
https://opentext.uleth.ca/apex-video/sec_power_series.html#ex_ps2
a power series?

The definition:
Definition 9.6.6: Radius and Interval of Convergence 
https://opentext.uleth.ca/apex-video/sec_power_series.html#def_radius_converge
sounds like it wants that term to only apply to power series, since it refers to the R in:
Theorem 9.6.4: Convergence of Power Series 
https://opentext.uleth.ca/apex-video/sec_power_series.html#thm_radius_converge

If all seems fine here then never mind. An alternative is to rephrase Theorem 9.6.7. Or maybe remove it. I think I understand that the goal is to convince students it's right to start looking at 
sum |a_n (x - c)^n|
Then APEX's version of the Ratio Test requires positive terms, so now you can use the Ratio Test on this other thing that is not a power series. I think there must be a concise but clear way to deliver that message.

APEXCalculus:
First, Alex's question led me to discover some errors that need fixing:
- Item 4's problem statement in Example 9.6.8 (
https://opentext.uleth.ca/apex-video/sec_power_series.html#ex_ps2) doesn't
match its solution; they address two different problems. The statement is
new; the solution matches the original text.
- In the same problem, the solution to part 2 introduces " n! " in the
next-to-last step erroneously.

Alex raises an issue that has gone overlooked. I've done some searching
through texts in my office to see what others did & what I was thinking at
the time of writing. Stewart and Larson/Edwards define the Ratio Test by
using the absolute value, hence skip this issue. I don't recall why I chose
to write the Ratio Test as requiring positive terms. Thomas has a similar
Ratio Test as Apex, then somewhat skirts the issue when finding the radius
of convergence.

I can easily be wrong about this, but it seems that finding the radius of
absolute convergence leaves open the possibility that there is a larger
interval of conditional convergence (beyond including just endpoints).
[...]

PR #265, opened 2023-08-04
More miscellaneous edits

sean-fitzpatrick:
This pull request takes care of a few more outstanding issues.

I am going to take a moment to see if I can do anything about #179 and then call it a week.

### DIFF
--- a/ptx/sec_FTC.ptx
+++ b/ptx/sec_FTC.ptx
@@ -2745,7 +2745,7 @@
             </pg-code>
             <statement>
               <p>
-                <m>F(x) = \ds\int_{<var name="$low"/>}^{<var name="$high"/>} <var name="$f"/>\, dt</m>
+                <m>F(x) = \ds\int_{<var name="$low"/>}^{<var name="$high"/>} (<var name="$f"/>)\, dt</m>
               </p>
               <p>
                 <var name="$D" width="30"/>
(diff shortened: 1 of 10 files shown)


sean-fitzpatrick:
I've added a commit that attempts to adjust the language related to power series, as discussed in #179 .

I'm fairly certain that both @APEXCalculus and @Alex-Jordan will want to suggest edits to what I've proposed.

sean-fitzpatrick:
I made some adjustments. I didn't add the last sentence ("The following theorem...") since we have replaced the theorem with a Key Idea.

APEXCalculus:
I think all this looks good. I can merge, but I will wait in case someone else wants to look it over first.

sean-fitzpatrick:
Since #179 was opened by @Alex-Jordan I'll wait to see if he's happy with the change here.

Alex-Jordan:
I think you should do what you and @APEXCalculus think best. Circumstances here are such that I'm not teaching calculus these days and I don't expect to for a while. And most (not all) of my colleagues are choosing to us either Stewart or OpenStax. So I think my role with APEX evolution can be more of a technical advising role now about WeBWorK and PreTeXt, and maybe you can tolerate my musings and observations every now and then like #179. But you two should steer things.

FWIW, I did read over the ratio test/ interval of convergence edits and I think it's good.

sean-fitzpatrick:
I still really want to know what it is that people are in OpenStax that I don't. Any time I've looked at their stuff, it seemed to me that whatever money they're making is not being reinvested in improving the books...

In any case, I'll go ahead and merge this.

When I'm at my computer (currently on phone) I'm going to post here with a WeBWorK problem that has me stumped.
Somewhat at random it will fail to build, always the same problem. But every time I re-run with debugging turned on, it builds without any problem.

PR #266, opened 2023-08-04
Runestone setup

sean-fitzpatrick:
I've had this work sitting on a branch for over a month, and never opened a pull request.
I guess maybe because I went on vacation.

In any case, this adds a Runestone target to `project.ptx`, and adds the necessary bits in the `docinfo`.


sean-fitzpatrick:
I think this is all set. The changes only affect the Runestone target (and some of the videos) so I think I can safely merge.

PR #269, opened 2023-08-07
Comment out review exercises

sean-fitzpatrick:
Only 15 sections (out of 97?) have review exercises.

For consistency, until we can do all, it's probably better to have none.


APEXCalculus:
15 out of 97 is a very low percentage!

PR #271, opened 2023-08-16
Some error fixes and adjustments

sean-fitzpatrick:
Fixes to a few problems I noticed while building PDF.

Also, I started adding xml:ids to asides (this is with the hope that we can eventually encode all of the placement of margin items in a CSV file that gets read in with our extra XSL -- @Alex-Jordan suggested this might be possible).


sean-fitzpatrick:
These are all minor typo and bug fixes, so I'm going to merge right away.

Issue #272, opened 2023-08-19
Incorrect conditions in curl theorem 


sean-fitzpatrick:
There's a mistake in [Theorem 15.3.18: Curl of Conservative Fields](https://opentext.uleth.ca/apex-calculus/sec_line_int_vf.html#thm_conservative_field_curl) .

The conditions state that the vector field should have components that are continuous on a simply connected domain. This should be continuously differentiable, not continuous. Otherwise there is no guarantee that the curl vanishes: the potential function has to be C^2.

A colleague pointed out the example f(x,y)=(x+y)|x+y|, which has gradient F(x,y)=(2|x+y|, 2|x+y|). This is a conservative vector field, but its curl isn't zero, since it's undefined on part of its domain.

sean-fitzpatrick:
My colleague was also confused by the statement that the vector field is conservative if and only if the curl is "0, or \vec 0, respectively".

The use of curl to refer to the scalar quantity Q_x-P_y for a 2d field is not common terminology, I think.

I am not suggesting that @APEXCalculus needs to remove this language, but I think there could be some clarification in this theorem?

APEXCalculus:
I don't know why the thm says continuous; that certainly doesn't imply the
ability to take a derivative as needed by the def. of curl.

As for the scalar computation of curl: I don't know what is common. It
seems that that is how I first learned the def. of curl in the plane, but a
look back at Kreyszig implies that I must have learned the scalar notion
elsewhere. It seems natural to me to have the curl of a 2D vector space be
a scalar and not a 3D vector, but I suppose "natural" is very subjective. I
impart meaning of the scalar to the 2D space, whereas others impart a
meaning of the 3D curl to the 2D space.

So the condition needs to change. I'll think about how to reword the thm to
make the second issue more clear.

sean-fitzpatrick:
I don't think there's any standard terminology in two dimensions. I did an undergrad in physics and learned calculus from Stewart. I only recall curl being used in 3d, and I don't think there was a word in 2d.

I'll look to see how other books state that theorem.
It might suffice to have a margin note with a reminder that in R2, we define curl as a scalar, but in R3, it's a vector.

sean-fitzpatrick:
Naturally, my colleague's email arrived about an hour after I had put this year's batch of PDFs up on the web.

Speaking of which: here is your version [in colour, for two-sided printing](https://opentext.uleth.ca/PDF/apex-color-print.pdf)

[...]

PR #273, opened 2023-08-19
More small edits

sean-fitzpatrick:
A few more adjustments noticed during PDF publication.

@APEXCalculus I've made a first pass at fixing the curl theorem menioned in #272.

I reverted the HTML style back to CRC. The problem with MathJax menus is still there, but hopefully will get fixed upstream. The default style just doesn't look as nice.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ APEX.pl
 # logs and error reports
 .error_schema.log
 cli.log
+vshifts.ods
--- a/ptx/sec_deriv_implicit.ptx
+++ b/ptx/sec_deriv_implicit.ptx
@@ -880,7 +880,7 @@
           Given <m>y=x^x</m>, use logarithmic differentiation to find <m>y'</m>.
         </p>
       </statement>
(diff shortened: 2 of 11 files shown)


APEXCalculus:
It hit me as I read the curl statement that the term 'continuously differentiable' is not defined in the text. (At least, my search did not turn up the term.) 

How about:
 Let <m>\vec F</m> be a vector field whose components have continuous derivatives
          on a simply connected domain <m>D</m> in the plane or in space.

sean-fitzpatrick:
Oh, good point. We don't use that language. Should we? I guess I use it constantly in the classroom and didn't even realize it's not in the book!

(I also use C^1, C^2, etc but we probably don't need to go that far.)

sean-fitzpatrick:
OK. Short term fix incoming.

sean-fitzpatrick:
Sorry, looks like my timing was off!
The last commit is the fix.

APEXCalculus:
No worries - merged!

PR #276, opened 2023-10-18
One fix, and some 2.0 compatibility things

sean-fitzpatrick:
On the last pull request I forgot that for TikZ images in a webwork problem, `$...$` needs to change to `\(...\)`.

I've also updated the requirements.txt file to the 2.0 version, which now updates automatically when pretext is updated.
I was told that the `devcontainer` file should be included to make things easier for anyone who wants to build on codespaces, so that's included.

I also updated the `project.ptx` file to the 2.0 syntax. There is one question with that; opinions from @Alex-Jordan or @APEXCalculus are welcome:

I've included targets for the versions of APEX that I build for our courses here. Previously I maintained a separate branch with a different `project.ptx` file that included them, but this became a big headache: every time I tried to update my branch from `main`, there was a merge conflict due to the different `project.ptx` files. This makes it a lot easier for me to build and publish the different versions of the book.

The downside is that for these builds, instead of `apex.ptx` I have other files (e.g. `apex-UL-standard.ptx`) that put the chapters in the order we use here. But those files (and the corresponding publication files) are not currently tracked by git, since they're specific to our institution.

Should I:

a) not include my build targets in the project file at all, and continue to use a different branch for my versions?
b) comment out those targets?
c) track the files those targets need, even though they may not be useful to others?
d) just leave things as they'll be after this pull request and not worry about some targets that will be missing files?

Let me know which is best, and I'll add one more commit here before merging.


sean-fitzpatrick:
Partially answered my own question: if files are missing for a target in `project.ptx`, you can't build, even if you aren't trying to build that target. I've commented out those targets for now, and can remove them entirely if that's preferred.

APEXCalculus:
I think I understand the issue. I think two things are important.
First: newbs to APEX pretext should be able to compile files as found in
the repo. They need not know to comment out lines in a file first.
Second: the guy (named Sean) who uses the APEX pretext repo the most should
be able to create the files he needs for his personal/institutional use
without headache.

I think option c) from the first post satisfies both needs.

If my understanding is correct, creating a project.ptx file with certain
targets commented out may also work. 99% of users won't uncomment those
targets, but Sean will when he creates the UL versions. If doing so isn't a
headache, then that is also a fine option.

I hope this helps.

sean-fitzpatrick:
Thanks Greg. The files I would need to commit are different chapter files, since that's where we do the organizing.

I could think about a better naming scheme before I commit them.

The main differences:

1. There are two main files, "standard" and "accelerated", which correspond to the two calculus sequences we teach here. The book is divided into parts representing the courses in each sequence.
2. We teach Taylor polynomials in Calc 1, so that section gets moved out of sequences and series and into applications of the derivative, changing the files for both of those chapters.
3. In both sequences we have a weird cutoff, where functions of several variables get introduced at the end of one course, and finished in the next, so that chapter gets split into two.
4. I have different publication file, but I have to review to see if those are strictly necessary.

It is also not a big deal to leave my targets commented out. Then if I try to build one of my versions, pretext says "I can't find that target", and I think, "oh yeah, I need to go uncomment that line".

My other solution would be to keep an untracked project file with a different name, and swap files every time. This is slightly more annoying but still manageable.

sean-fitzpatrick:
Yep! That's exactly what I have now. It just isn't tracked on GitHub because I didn't think the rest of you would want those files cluttering things up.

APEXCalculus:
I doubt it's that much clutter. You use the repo so much more than others
that it seems worth letting you have some time-saving features.

PR #278, opened 2024-01-26
Update project.ptx to use intended defaults

StevenClontz:
Right now targets are building in `output/output/TARGETNAME` (using the default `output/` project output directory, and an `output/TARGETNAME` target output directory). I assume the intention is to build in `output/TARGETNAME`, which is the default, so I've removed the unnecessary configurations.

I've also fixed the Runestone target to build into `published/APEX` rather than `output/published/APEX`.

### DIFF
--- a/project.ptx
+++ b/project.ptx
@@ -1,35 +1,30 @@
 <?xml version="1.0" encoding="utf-8"?>
-<!--
-    This file provides the overall configuration for your PreTeXt
-    project. To edit the content of your document, open `source/main.ptx`
-    (default location).
--->
 <project ptx-version="2" source="ptx" asy-method="local">
   <targets>
-    <target name="html" format="html" source="apex.ptx" publication="publication.ptx" output-dir="output/html">
+    <target name="html" format="html" source="apex.ptx" publication="publication.ptx">
       <stringparams key="html.annotation" value="hypothesis"/>
     </target>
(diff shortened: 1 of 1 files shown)


sean-fitzpatrick:
Was there a change to the CLI? I thought I had finally figured out how to get things into the right places.

I know we are supposed to use `source` and not `ptx`, but that folder name predates the CLI by a couple of years.

StevenClontz:
There have been (and will be) no changes in `<project ptx-version="2"/>` schema since it released.

Default location for all project source files is `source`, which you've correctly configured to `ptx`. You could rename the `ptx` directory to `source` and remove `project[@source="ptx"]` if you wanted, but this is fine.

You're correctly using the default location for all project publication files, which is `publication`.

I think you intend to use the default location for all project output files, which is `output`. The default location for each target is `TARGETNAME`, but you've configured `output/TARGETNAME`, which builds things into `output/output/TARGETNAME`. Deleting all the unnecessary `output-dir`s as proposed gets you the behavior I think you intend. (Except for Runestone, which will now build into `output/../published/APEX = published/APEX` as necessary.)

sean-fitzpatrick:
OK, I see why I was mixed up: I've been working on a fork since September where I've already made this change. I haven't built from the main repo since summer; nor have I sent any pull requests from the active fork. (We've been working on writing image descriptions.)

StevenClontz:
FYI - this came up because I believe @bnmnetp is building your Runestone book from this branch. Thanks for merging.

PR #295, opened 2024-05-08
Ensure consistent/current xref syntax

sean-fitzpatrick:
We've had a hodgepodge of syntax usage for cross references for quite awhile.

The original conversion used syntax like `<xref ref="eg_foo">Example</xref>` for pretty much all cross-references; I think this was the standard at the time.

I've updated most of these to use the syntax `<xref ref="eg_foo"/>`. I think this is more future-proof, and also has the advantage that the link text is inherited automatically. (For example, if a figure becomes a table, or vice-versa, the link text changes accordingly, with no action required on our part.)

Exercise labelling was a mixed bag. There were a lot of things like "In Exercises `<xref ref="ex_first"></xref> <ndash/> <xref ref="ex_last"></xref>`" that I've updated as "In `<xref first="ex_first" last="ex_last" text="local">Exercises</xref>`"

This should now be consistent across the book, although it's possible I missed one or two. The link text now looks like "Exercises 13 -- 18". (The `text="local"` means we use the exercise number, rather than "Exercises 4.5.13 -- 4.5.18". I think the former makes more sense, since these exercise references are internal to a section.

For "Part", "Item", "Step", "Rule" I kept the supplied text, with local numbering. I used global numbering (including chapter and section number) when referring to definitions, theorems, etc.

One note: for equations I went with the syntax `<xref ref="eq_foo" text="global">Equation</xref>`. This makes "Equation" part of the cross-reference link. We were about 50/50 between doing this and writing "Equation `<xref ref="eq_foo"/>`", which leaves "Equation" out of the link text. And there were a few instances of just `<xref ref="eq_foo"/>`, which refers to the equation by number only, without "Equation".

If a different style is preferred for equations, it shouldn't be hard to adjust.

### DIFF
--- a/ptx/chapter_app_of_int.ptx
+++ b/ptx/chapter_app_of_int.ptx
@@ -4,7 +4,7 @@
   <title>Applications of Integration</title>
   <introduction>
     <p>
-      We begin this chapter with a reminder of a few key concepts from <xref ref="chapter_integration">Chapter</xref>.
+      We begin this chapter with a reminder of a few key concepts from <xref ref="chapter_integration"/>.
       Let <m>f</m> be a continuous function on <m>[a,b]</m> which is partitioned into <m>n</m> equally spaced subintervals as
       <me>
         a=x_0 \lt  x_1 \lt  \cdots \lt  x_n\lt x_{n}=b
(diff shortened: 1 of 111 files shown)


sean-fitzpatrick:
One note: cross-references in the section on trig substitution were  a mess. Things like `<xref ref="idea_foo">Key Idea</xref>(b)` were hard-coded, but there was no part (b)! This is now fixed.

I also changed one image in this section: I rotated the label on the last of the three reference triangles, so that they all have similar aspect ratios, and sizing is more consistent.

PR #296, opened 2024-05-09
Move video solutions to second spot

sean-fitzpatrick:
When I put the video solutions into their own solution knowl, they ended up as the first solution, because I could do this with an easy find+replace (and some regex).

But video solutions should not come first. I finally did the manual work required to put the video solutions after the text solutions, so that the text solution comes first.

So in HTML, instead of seeing:
> Solution 1 (Video Solution)
> Solution 2

we will see

> Solution 1
> Solution 2 (Video Solution)

This will affect placement of QR codes in the margin for PDF that includes video links, but I suppose that's my problem.

I also cleaned up some trailing whitespace.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ vshifts.ods
 # CLI bloat
 codechat_config.yaml
 .devcontainer.json
+.github/workflows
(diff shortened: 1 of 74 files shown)


APEXCalculus:
Sheesh. That seems like a lot of work.

PR #297, opened 2024-05-13
Move image xml:id to label on latex-image/asymptote

sean-fitzpatrick:
This is a technical change under the hood to ensure forward-compatibility with some changes to the PreTeXt machinery.
In particular: we had been using `xml:id` on image elements as a way to ensure that generated images had unique file names.
This was, at one time, best practice.

The updated philosophy is that `xml:id` should only be used on things that can be the target of a cross-reference (e.g. figure, but not image). File names should come from a `label` on the `latex-image` or `asymptote` element inside the image.
Chapters and sections should also get labels, since these each get their own html file.

@Alex-Jordan the style sheet that Rob and I put together to do the changes for images does a bit of additional clean-up, which includes changing all instances of `>` to `&gt;`. I didn't encounter any issues building webwork representations, but please let me know if this raises any red flags. I could change some of these back if necessary.

I'm seeing a few generated html files with funny names that I need to check on (`apex-calculus-2`, `apex-calculus-2-2`, etc). I think this is a sign that I either missed some labels, or labelled some things incorrectly. I'll add a comment here once I've looked into this.


sean-fitzpatrick:
I've fixed the funny html file names. They were coming from pieces like the front matter, back matter, colophon, etc.

APEXCalculus:
I've been merging Sean's recent work knowing that he's checking his own work as he goes along.
Since this once calls out @Alex-Jordan , I'll give him a chance to provide a thumbs-up.

Alex-Jordan:
Thumbs up!

sean-fitzpatrick:
OK! I'll go ahead and merge.

PR #298, opened 2024-05-14
Errata fixes

sean-fitzpatrick:
Various changes to address some items in the issues:

- fixes to some webwork code that was giving incorrect answers
- a typo fix
- add an aside noting the difference between curl in 2d and curl in 3d
- add definition of continuously differentiable function

@APEXCalculus I am not sure if the last item is how you would proceed.
I did a search for the phrase "continuously differentiable": it shows up several times in the sections I added for the Lethbridge versions, which is probably not a big deal.

But the phrase is also used a few times in the two sections on line integrals.

The other way I can proceed here is to not add this definition, and in the line integral sections, replace the occurrences of "continuously differentiable" with "has continuous partial derivatives".

Let me know which option you'd prefer.


sean-fitzpatrick:
The WeBWorK code that I thought I had fixed is not fixed. I've posted on the PreTeXt forums to see if @Alex-Jordan can help: the webwork code he wrote originally doesn't work, and neither does my attempt at fixing it -- there is a problem with the "show your work" elements.

(I am not sure those can ever work inside the book, so one option may be to consider removing all the webwork essay-answer stuff and replacing it with something else.)

sean-fitzpatrick:
I have added changes to the scatterplot images, so that we use squares for the second sequence.
Then I had to adjust a bit: if you want the circle on top of the square, you need to plot it second.

I think the webwork stuff is settled now. The original code was fine, but there's a bug in a macro library.
And I needed to change a setting, which has the useful side effect of causing webwork to ignore the buggy part.

Rebuilding once more to check things over.

sean-fitzpatrick:
OK, I think the images are good to go now.

@APEXCalculus the main thing I need from you before this can be merged: would you rather add a definition of what it means for a function of two variables to be continuously differentiable, or edit the few occurrences of this terminology?

APEXCalculus:
I think the definition addition is good. Having that term defined will be useful for some who encounter it elsewhere.

sean-fitzpatrick:
Sounds good! I suppose we could be more systematic about usage: in other places there are theorems with conditions like "has continuous partial derivatives" that could be replaced by "is continuously differentiable" but I don't think it's a priority. 

I made the definition for functions of two variables, in what I thought was the most natural place. I didn't repeat it for functions of three variables, vector fields, etc. but I think it's clear how to extend.

PR #305, opened 2024-06-21
Automate PDF production

sean-fitzpatrick:
This adds several attributes used by the LaTeX conversion (and some new XSL to process them).

It should no longer be necessary to manually edit the `.tex` file produced by PreTeXt: the adjustments needed for vertical placement of margin figures are now in the PreTeXt source.

There are also other bits to deal with a small number of boxes that aren't wide enough, or that need to be shifted to the left for even pages in two-sided printing, as well as a few automatic page breaks.

I also cleaned up some old ptx markup for some of the WeBWorK exercises so they look better in print.

You *should* be able to directly build to PDF, but it's probably a good idea to still build the `.tex` file first, and then process it. (If you go straight to PDF, it doesn't keep the .tex file, and then you have to start over if you notice some small thing that needs to be changed.)

To build `.tex` (and all the auxilliary files), do `pretext build latex` using the CLI, or `pretext build latex -g` if you want to force generation of WeBWorK, Asymptote, etc., and then process the resulting file with `xelatex`.

The `project.ptx` file lists all the various versions of PDF that can be built.


sean-fitzpatrick:
@APEXCalculus you should be able to build PDF on Windows if you have LaTeX+Asymptote installed, and you install Python, and then `pip install pretext`.

You will probably be able to see places where you want to do some fine-tuning, and you can now do that in the PreTeXt source by changing the value of the `vshift` attribute on figures and margin notes.

(Note that for the versions without QR codes for the videos, things are sometimes more spaced out than you'd otherwise do, to leave room for the QR codes in the versions that have them.)

PR #306, opened 2024-06-25
A few small fixes to previous PR

sean-fitzpatrick:
Fixing a couple of small items that I missed in the last pull request:

- I accidentally deleted a couple of necessary line breaks around minipages (with two-sided printing, there are some over-wide boxes that I need to shift over, or I'd vanquish the minipage altogether...)
- I also decided that we needed `\medskip`, not `\smallskip` around those minipages
- In the Taylor polynomial section, I mixed up the "early" and "late" components in a couple of places. (My versions are "early", since we do Taylor polynomials as part of Chapter 4.)
- I also noticed some places where there was a space between `\left` and `\right` and the delimiter. This doesn't seem to be a problem but I fixed it anyway since it seemed wrong.

### DIFF
--- a/ptx/sec_Graphical_Numerical.ptx
+++ b/ptx/sec_Graphical_Numerical.ptx
@@ -284,8 +284,8 @@
         </p>
         <p>
           <ol marker="(a)">
-            <li><m>y = C \left ( 1 + \ln(x) \right )^2</m></li>
-            <li><m>y = \left ( \frac{1}{3}x + \frac{C}{\sqrt{x}} \right )^2</m></li>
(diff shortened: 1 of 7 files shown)


sean-fitzpatrick:
@APEXCalculus if you have time to build one of the PDF targets and go through it to see if everything looks ok (and tell me if it doesn't) that would be great

APEXCalculus:
I'll see what I can do on Friday; I am traveling until then.
I haven't had much success building using Codespaces. I'll look through the
guides and our old emails to see if I get good clues as to how to make it
all work.
Will keep you posted.

PR #308, opened 2024-07-19
Add support for inserting 'enlargethispage' and some small adjustments

sean-fitzpatrick:
As the title says. I thought there was a problem with duplicate ids in one place, but this turned out to be a bug in the CLI.

The fix to add some space after a sidebyside is in a pull request to the main pretext repo.


sean-fitzpatrick:
I also removed some redundant ids on video tags, which is why you're seeing a lot of files changed.

PR #311, opened 2024-07-31
Edits in Chapters 1-4

APEXCalculus:
Mostly the edits are the shifts in figures.
Other edits are made too, changing some wording in places. (Some key ideas, definitions, asides are edited.)


sean-fitzpatrick:
I'm having some trouble tracking all of the changes.

- In the main `apex.ptx` file, better to not commit the termporary changes to comment out the chapters you're not building.
- There are a lot of whitespace changes (e.g. in some of the chapter files, some of the `<xi:include>` lines have been indented for some reason.
- Later on there are a few things that look like large changes, but I suspect are just changes in indentation. If you can stick to the existing indentation scheme, it'll make it easier for me to follow along
- In a lot of places you've made a small change (like changing `<dl>` to `<ol>`) by commenting out the original, and then adding a copy of it below that has the changes. Unless there is a reason you want to keep the original in a comment, I think it's a lot easier to see what's going on if you just make the chnage.

(Also, the copied versions had a lot of new, extraneous whitespace that made it hard to compare.)

APEXCalculus:
Yes, I am sure this pull request is a mess, much of it due to my status as a git noob. I can redo it in a cleaner way if necessary. Problems I foresaw once I committed:
1. There is too much in this one pull request. I should do one per chapter. Or per section if better.
2. Yes, apex.ptx is changed and I didn't mean to commit that. 
3. I made changes that I wasn't sure were good, so I commented out old stuff in case we wanted to refer back to the original. I forgot that git handles that kind of issue well; I am too used to doing things on my own.

Is it better if this pull request is cancelled and I:
- commit in smaller chunks
- only commit the sections I am editing (and not apex.ptx & chapters, etc.)
- delete old content and not comment it out, and
- try to manage whitespace better?

Sorry for the trouble.

sean-fitzpatrick:
I can't complain about large pull requests, considering some of the ones I have sent!

Only start over if it's easier for you.

Otherwise, if you can revert the apex.ptx and chapter files, and clean up some of the whitespace, I should be able to work through the rest.

APEXCalculus:
I didn't officially "revert" apex.ptx and chapter files back to the original, but I did restore them to their original state and re-committed them in commit "fixes apex.ptx and chptrs 1 & 2": (https://github.com/APEXCalculus/APEXCalculusPTX/commit/7be3a907a38f699990e098607f7ef9f32144885e)

I edited sections to remove most content that I commented out in "fixes sections":
(https://github.com/APEXCalculus/APEXCalculusPTX/commit/f2add7028f2d90afbf6f2c7cd83c8328c9198a72)

If I need to adopt a more git-official method, let me know.

One of the most confusing forms of edits is where certain figures needed to be moved before/after certain paragraphs so that they appear on the proper page. So if it looks like I deleted figures, it's because of that.

sean-fitzpatrick:
I added a few other changes to the section on basic derivative rules.

The pagebreak element should probably not go inside a paragraph, etc.

A couple of questions for you about changes to the placement of figures, changes to vshifts, etc.:

1. When you move a figure, will the new location still make sense if the figure is inline? We only put figures in the margin for PDF output. Probably 95% of students (or more) use the HTML version of the book. Very few are interested in PDF. We get only a handful of orders for print copies every year in our bookstore. So I think it's important not to change anything that compromises the HTML output.
2. Are you using the verison with or without the QR codes for videos? If the new locations overlap with the QR codes, and there isn't room for me to move the QR codes so they fit, I'll have to do the figure placement all over again.

(I could use versioning for this, and have one set of vshifts for videos, and one for no videos, but it would require duplicating half the figures in the book, so I can put one in the `video` component, with one vshift, and another in the `novideo` component, with a different vshift.)

APEXCalculus:
0. I wasn't sure where the pagebreak command should go. It seemed to me
that this command and the enlargethispage command acted differently; one
seemed to work better inside a paragraph and one didn't. But I could be
wrong, and I'm fine leaving them both outside of paragraphs.

1. Ugh, no, I hadn't thought of the inline nature of the graphics. I'll
have to go back and look at instances where I moved them. I *think* I only
moved them from one side of a paragraph/thm/def to the other, but I could
be wrong. This was only done when there was an automatic
pagebreak occurring, and maybe I can work around it by inserting a
pagebreak command.

2. Ugh, no, I wasn't looking at the version with the QR codes. I can
restart from Chptr 1 and put the codes in and see how things fit around
them.

This is for sure a stress-test of PreTeXt. There are several competing
desires for content and its placement. My school may be backward, but few
of my students (and colleagues) use the website and most use a printed
text, though some prefer the pdf. That's why my focus is always on the
printed version.

sean-fitzpatrick:
I think a page break should go between paragraphs, the same as it does in LaTeX. 

If the enlarge page command works better elsewhere, that's ok. (In LaTeX I don't think I've ever put it in without line breaks on either side of it.)

Have your students been given the option of the online version of the book? I can understand preferring a hard copy, but I'd be surprised if anyone chose electronic PDF over HTML.

I think people in some places are being pushed to move to HTML to comply with accessibility requirements. But you also get nice perks like interactive features. 
(The built in search works really well, too.)

ptx/sec_deriv_interpret.ptx
```diff
@@ -1137,10 +1141,12 @@
         </webwork>
       </exercise>
 
+			<p><pagebreak-latex/></p>
```
sean-fitzpatrick, line 1144: There are three consecutive pagebreaks here. Shouldn't there be at most one?

I think you need to remove the ones on lines 1127 and 1131.

You could also try editing the .tex file, and adding `\nopagebreak[4]` at the end of the introduction to the exercise group (immediately after the text "Identify which function is the derivative of the other.").

Eventually this will be added automatically, but there is still some ongoing discussion about it.

ptx/sec_differentials.ptx
```diff
@@ -159,20 +159,22 @@
...
       <p>
-        The function <m>\ell(x)</m> is often referred to as the <term>linearization</term>,
-        or <term>linear approximation</term> of <m>f</m> at <m>c</m>.
-        It is the linear function that best approximates the value of <m>f(x)</m> when <m>x</m> is close to <m>c</m>.
+        Let <m>f</m> be differentiable on an interval <m>I</m> containing <m>c</m>. The function <m>\ell(x) = \fp(c)(x-c)+f(c)</m> is the <term>linearization</term>,
+        or <term>linear approximation</term>, of <m>f</m> at <m>c</m>.
+        <!--
+				It is the linear function that best approximates the value of <m>f(x)</m> when <m>x</m> is close to <m>c</m>.
+				-->
         <idx><h>linearization</h></idx>
```
sean-fitzpatrick, line 178: By removing this line, aren't you deleting a reference to the only reason one would bother using the linearization? The whole point is not that it is _**a**_ linear approximation, but the _**best**_ linear approximation.

This could be moved out of the definition, perhaps, but doesn't it need to be mentioned?

Is not the *entire point* of differential calculus finding the best linear approximations to nonlinear functions?

ptx/sec_differentials.ptx
```diff
@@ -408,15 +410,26 @@
...
-    <title>PID controllers</title>
+  <aside vshift="3">
+    <title>Applications of Differentials</title>
     <p>
-      Another place differentials are used is in a PID controller,
+      Differentials are used within <q>Proportional Integral Derivative</q> (PID) controllers which use both integral and differential calculus to accurately control a process.
+		</p>
+		<p>
+			For instance, consider the task of steering a self--driving car. If the vehicle drifts to the left, how much should the wheel be turned to the right to correct the path? Too little and the problem may not be corrected at all; too much and the car will overshoot its target and extreme corrective action will need to be taken. PID controllers consider the <em>rate</em> at which the drift and corrective measures take effect.
+		</p>
```
sean-fitzpatrick, line 420: The `--` for a hyphen is LaTeX code, but it is not valid PreTeXt. You want to use <ndash/>.

sean-fitzpatrick:
I made a few changes that were easy to make myself. An earlier comment explains how to fix the whitespace.
There are a few inline comments for things that need your attention.

Other than that, I think this is OK, pending confirmation that the changes are compatible with the HTML output and including video QR codes.

ptx/sec_deriv_interpret.ptx
APEXCalculus: I didn't intend for all those to stay. I think I was trying to put them in as many places as possible to see what would actually have the desired effect.

ptx/sec_differentials.ptx
APEXCalculus: Use <ndash/>?

APEXCalculus: Ha! Github does like writing the pretext code. Use < n d a s h / >?  (Without spaces - Github turns that command into a space.)

APEXCalculus:
I fixed my tab / 2 spaces issue -- I think. We'll see how it looks going forward.

ptx/sec_graph_sketch.ptx
```diff
@@ -1009,7 +1009,7 @@
     In <xref ref="fig_sage_sinx"/>,
     two graph of <m>y=\sin(x)</m> is given,
     generated by <pubtitle>Sage</pubtitle> and <pubtitle>Mathematica</pubtitle>.
-    The small points represent each of the places where each <init>CAS</init> sampled the function.
+    The small points represent each of the places where each <!-- <init>CAS</init> --> CAS sampled the function.
     Notice how at the <q>bends</q>
```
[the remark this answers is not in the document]
APEXCalculus: I probably fixed a problem in a wrong way. 
For some reason, that markup displays CAS as "cas" in the pdf. I supposed the way PreTeXt handles it should change, right?  This also appears in Example 3.5.8, though I missed fixing it there.
At the same time, I wasn't sure of the point of the init markup. My brief search for that markup just led me to running "pretext init" which wasn't helpful.

ptx/sec_differentials.ptx
sean-fitzpatrick: You can put pretext code between back ticks and it works OK. Which I clearly forgot to do!

So yes, you would do `self<ndash/>driving`.

ptx/sec_graph_sketch.ptx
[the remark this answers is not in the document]
sean-fitzpatrick: Ahh.... I think the problem is that the Carlito font set doesn't include small caps. I will ask if there is a good way to work around this.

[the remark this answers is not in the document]
sean-fitzpatrick: There is a [font based on Carlito](https://codeberg.org/yomli/latte-fonts) that includes small caps, but I didn't figure out how to get LaTeX to use it.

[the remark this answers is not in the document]
sean-fitzpatrick: I think we should keep markup like `init` because it will produce semantic tags in HTML, which is useful for those needing a screen reader. (It should know to pronounce each letter rather than trying to make it a word.)

Options:
1. Override templates in LaTeX so that small caps are regular caps
2. Change to a font that supports small caps
3. Keep the current font, but use an alternate font for small caps

Let me know what you prefer.

ptx/sec_deriv_prodquot.ptx
```diff
@@ -156,15 +156,15 @@
...
       <p>
         Adding <m>0</m> in some clever form is a common mathematical proof technique.
       </p>
     </aside>
     
     <p>
       <md>
-        <mrow>\amp\lzoo{x}{f(x)g(x)} = \lim_{h\to0} \frac{f(x+h)g(x+h)-f(x)g(x)}{h}</mrow>
+        <mrow>\lzoo{x}{f(x)g(x)} \amp= \lim_{h\to0} \frac{f(x+h)g(x+h)-f(x)g(x)}{h}</mrow>
         <mrow>\amp \text{ (now add 0 to the numerator) }</mrow>
```
sean-fitzpatrick, line 168: Moving the alignment tab here is going to cause a bunch of overfull hboxes.

That will be a problem if this ends up on an odd-numbered page, I think.
And it means that the equations won't fit in HTML: a horizontal slider will be needed.

sean-fitzpatrick: or maybe even pages, not odd, since there's nothing in the margin there.

Alex-Jordan:
About the font/smallcaps/init issue. If there is nothing better to do for PTX at large at this time, you could use the custom XSL that this project already has to change how an `init` is translated. And just not use small caps.

sean-fitzpatrick:
@APEXCalculus I will have time this month to return to this. If you would like to connect one day to go through everything, let me know!
I will try to catalogue the changes here, and maybe redo them on a fresh branch -- GitHub is showing some conflicts.

Regarding placement of margin figures in the PDF: why don't we get figures and asides where you want them, and then it'll be my problem to work out how to fit the video QR codes around them.

ptx/sec_differentials.ptx
sean-fitzpatrick: Actually I think an em-dash is too long here. An en-dash would be better.

ptx/sec_graph_sketch.ptx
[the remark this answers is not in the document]
sean-fitzpatrick: I am going to leave the markup in place, and in the extra XSL, I can kill the LaTeX styling that is causing this to appear lower case.

ptx/sec_limit_continuity.ptx
```diff
@@ -234,34 +264,6 @@
...
-          </li>
-          <li>
-            <p>
-              The limit <m>\lim\limits_{x\to c} f(x) = f(c)</m> for all values of
-              <m>c</m> where the limit exist,
-              since each step consists of just a line.
-            </p>
-          </li>
-        </ol>
-      </p>
```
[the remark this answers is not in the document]
APEXCalculus: I made that change 8+ months ago. I no longer remember why. :)
I was working on figures in the margins, so it would have moved in order to fit on the page better. For sure, it belongs with the question statement, not just the solution. I'd ignore this change and figure out the print version placement later.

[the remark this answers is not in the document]
sean-fitzpatrick: My guess was that you wanted the image to move to another page :-)

I'm going through and adding all the non-cosmetic changes on a branch. We can get those in, build the PDF, and then worry about image placement.

ptx/sec_graph_sketch.ptx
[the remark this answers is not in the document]
sean-fitzpatrick: I have updated the XSL so that no formatting is applied to initialization, abbreviation, and acronym for PDF.

ptx/sec_related_rates.ptx
```diff
@@ -1116,15 +1116,15 @@
           </pg-code>
           <introduction>
             <p>
-              An F-22 aircraft is flying at
+              An F-22 aircraft is flying at <m><var name="$spd"/>\,\text{mph}</m>
```
sean-fitzpatrick, line 1119: The webwork variable `$spd` is already a number with units. This will result in a display like 530 mph mph.

sean-fitzpatrick:
This has been replaced by more recent pull requests. Since this is pretty stale now, I think we can close it.

sean-fitzpatrick:
This pull request is now very out of date, and obsolete, I think. I'm going to close it.

APEXCalculus:
Thanks.

PR #316, opened 2024-09-30
Update sec_antider.ptx

APEXCalculus:
Possible correction to antidiff. property.

[Theorem 5.1.6](https://opentext.uleth.ca/apex-calculus/sec_antider.html#thm_indef_alg-2-2) states  \int c \cdot f(x) \, dx = c \cdot \int f(x) \, dx.
While this is identical to language in other texts, I got a suggestion that we should add "(c \neq 0)" as a condition.

The reason (which you can figure on your own) is that if c=0, integrating 0 leads to a +C, whereas integrating first then multiplying by 0 means you just get 0. 

I'm wondering what others think of this suggestion. I'm lean toward making the change. 
Yes, I think they have a point. 
I also think the expression "c \cdot \int f(x)\,dx" is also playing a bit loose with set notation, and this may be an unimportant nitpick. But if it gets students thinking about what's going on, it may create a bit of a learning moment.

Interested in any discussion before someone merges.

### DIFF
--- a/ptx/sec_antider.ptx
+++ b/ptx/sec_antider.ptx
@@ -374,7 +374,7 @@
         </p>
         <p>
           <md>
-            <mrow>\amp\lzoo{x}{cf(x)}=c\cdot\fp(x)\amp\amp\int c\cdot f(x)\,dx= c\cdot \int f(x)\,dx</mrow>
+            <mrow>\amp\lzoo{x}{cf(x)}=c\cdot\fp(x)\amp\amp\int c\cdot f(x)\,dx= c\cdot \int f(x)\,dx\quad(c\neq 0)</mrow>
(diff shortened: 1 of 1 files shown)


sean-fitzpatrick:
I'm fine with this change. Since we define the indefinite integral as a set, the notation is a little awkward, but in practice I think everyone thinks of the indefinite integral as a representative of this set, with the value of C left unspoken.

Probably the best option (mathematically, not pedagocially) is to think of this set as an equivalence class: we have many examples throughout mathematics of operations on equivalence classes.

sean-fitzpatrick:
@APEXCalculus I got busy and forgot about this. I think you can go ahead and merge unless further discussion is needed.

PR #317, opened 2024-10-01
WeBWorK errors and typos

sean-fitzpatrick:
The last question in Section 2.3 (basic derivative rules) was coded incorrectly: all of the other problems in that section expect you to include the "y=" part of the equation for the tangent line. This ensures that the last question behaves the same way.

An easy fix, but I think it might be worthwhile to keep this pull request open until the end of the term, in case my students find other errors in the exercises.


sean-fitzpatrick:
I've fixed another incorrect answer calculation in Section 5.1 noted in #323

sean-fitzpatrick:
I'm going to merge this one to get that question working.

PR #326, opened 2025-02-19
Some collected housekeeping changes

sean-fitzpatrick:
A typo fix, a correction to a couple of webwork problems, and an update to the frontmatter, since the syntax for that has changed.


sean-fitzpatrick:
Nothing controversial here, so I'm going to break with protocol and accept my own PR.

PR #335, opened 2025-05-12
Non-figure edits from PR #311

sean-fitzpatrick:
This contains (I hope) all of the edits in #311 that were not related to placement of figures, page breaks, etc.


sean-fitzpatrick:
Note that I snuck in one edit of my own: in most of the book, we use sin(x), ln(x), etc. but there were some places with sin x, ln x, etc.

So I've added more parentheses. I'm sure a few others escaped me.

PR #338, opened 2025-05-26
Make Terms and Concepts interactive

sean-fitzpatrick:
This is an update to all of the "Terms and Concepts" questions in the textbook.
- All WeBWorK true/false and multiple choice are now using the PreTeXt infrastructure for this
- Any other true/false and multiple choice that weren't interactive are now
- For questions that ask students to "explain" or "describe", there is a `response` box. This will only show up when the book is used on a Runestone server
- All fill-in-the-blank are now dynamic
- There were a couple of fill-in-the-blank questions that involved math formulas. I couldn't make these dynamic fill-in, but I re-wrote them as "horizontal parsons": students take the pieces of the formulas, and rearrange them so that they're correct.


sean-fitzpatrick:
Update: the markup for the two horizontal parsons problems is fine, but PreTeXt hasn't "caught up" yet.

There is some Javascript coming from the Runestone side that needs to be updated to ge the evaluation working correctly for mathematical expressions.

APEXCalculus:
I checked the Ratio/Root test section and I think it looks good. 
Not sure about Merging - it seems there are still things you are working on? Or would you like it Merged and then you'll sort out the difficulties later?

sean-fitzpatrick:
This is a case where I think the source is settled, but the software is not.

Here is one of two problems that I coded as "horizontal Parsons":

https://opentext.uleth.ca/apex-calculus/sec_curvature.html#TaC-curvature-5

This was a fill-in-the-blank problem, but there's no good infrastructure to have students enter a math symbol like kappa.

So we put all the symbols in blocks, throw in some distractors, and have students assemble the correct equation. 

The problem is that right now, the correct equation is marked wrong.
This is an issue with some JavaScript on the backend that needs to be fixed.
Once that happens, the question should be fine.

PR #344, opened 2025-06-10
Finish all the image descriptions

sean-fitzpatrick:
I've finished writing descriptions for images that weren't done by the students I had working for me last year.
(Or was that two years ago?)

There are also some adjustments I made to images as I was staring at them, trying to decide how to describe them.


sean-fitzpatrick:
I don't think this is necessary for any of the images that use `pgfplots` (which is most of them).
Just the images that use pure TikZ.

PR #345, opened 2025-06-12
Adjust image scaling to improve font size in SVG

sean-fitzpatrick:
There were a number of images in the book that didn't look right due to scaling issues.

All of them are TikZ images that do not use PGF. There are [some instructions in the PreTeXt Guide](https://pretextbook.org/doc/guide/html/topic-images.html#topic-images-source-5) on how to calculate image scaling so that the text in the images matches the surrounding text.

I don't think I got this to work perfectly across the board, but certainly some of them are much better.
(The worst ones were in the sections on applications -- related rates, fluids, work, etc.)

I've updated all the [HTML versions](https://opentext.uleth.ca/calculus.html) in case you want to look things over,
and I've attached the PDF.

[apex.pdf](https://github.com/user-attachments/files/20716091/apex.pdf)


sean-fitzpatrick:
I'm realizing now that exercises are tricky: in PDF we reset the page geometry for exercises.
Text width is 320 pt in the main text, but I think it's 392 pt for exercises.
But there is no such change in geometry for HTML. That means an exercise image that looks OK in HTML might appear to have oversized text in PDF.

I try to optimize for PDF, and live with text that appears too small in HTML. I think that looks better than text that looks too big in PDF.

APEXCalculus:
Is this ready to be merged, or still a work in progress?
I did see the fonts in the exercises, and yes, they were too big. Probably "too small" is preferable to "too big," as you suggest.

sean-fitzpatrick:
I'm going to build the PDF again with the last set of adjustments. I'll post it here for you to look over.
If it looks OK to you, we can merge. (It's still not perfect, but it's a heck of a lot better than it was.)

sean-fitzpatrick:
[apex.pdf](https://github.com/user-attachments/files/20801960/apex.pdf)

Here is the most recent build. I had to fiddle with some of the "applied" images.
Scaling is easy when the TikZ code makes the size of the image (in centimetres) clear.

Once there are multiple scopes in play, units are in points, etc., I ended up doing a bit of guessing and checking.

If it looks good to you, you can merge. (I would check, in particular, sections like related rates, work, and fluid force.)

sean-fitzpatrick:
I am running an experiment with the templates we use to produce PDF.

Right now, margin images are hard-coded with `left-margin=0.02`, `width=0.96`, and `right-margin=0.02`. That width is relative to the marginpar width. This works great for 99% of images that move to the margins.

But some of the images shouldn't be 96% of the marginpar width, like those ones in Section 7.6.
(The reason the cylinder in 7.5 looks good and the the one in 7.6 looks bad is that the image in 7.5.9 has a bunch of surrounding text that increases the overall image width.)

So I am going to try to set up a calculation so that if a figure moves to the margin, its width is twice the authored width (2*47%=94%, which is slightly less than the hard-coded 96%, but I think it should be OK). Then we add other calculations to adjust the left and right margins accordingly.

So, for example, if we author `<image width="33%">`, we should get a width that is 66% of the marginpar width, with 17% margins on either side.

sean-fitzpatrick:
There is a catch... If the authored width is greater than 50%, then this sets the width in the .tex file to a value that is greater than 100%, with negative values for the margins. Not good...

But there were only two figures in the book that were set to go into the margins that had a width greater than 100%.
One is the cycloid in Section 12.1, and the other is the diagram in Section 9.2 that shows the procedure for plotting a parametric curve.

My thoughts on these are that perhaps they don't need to go into the margin. Both are naturally wide images that become hard to read if they are scaled too small.

APEXCalculus:
I'd say the images from 9.2 and 12.1 you mention shouldn't be in the margin. Yes, they are too wide for that. 
So does that mean your calculation workaround is ok?

sean-fitzpatrick:
All right, let's give this a try. There might be one or two oversights on my part (I caught two images in Section 2.1 that didn't have widths on them! I don't know how they still looked ok in HTML.)
[apex.pdf](https://github.com/user-attachments/files/20824240/apex.pdf)

I think we are almost there.

APEXCalculus:
Looks good to me! Much better. 
I'll merge (or you can) unless there are more tweaks you are working on.

sean-fitzpatrick:
I think you can merge! Thanks for checking everything over.

PR #361, opened 2025-07-25
Chapter 1 Placements

APEXCalculus:
Figure placements for chapter 1.
Added figure reference to sec_limit_continuity, else figure appears without any reference to it.

### DIFF
--- a/ptx/sec_deriv_basic_rules.ptx
+++ b/ptx/sec_deriv_basic_rules.ptx
@@ -241,7 +241,7 @@
         </ol>
       </p>
 
-      <figure xml:id="fig_xcubedwithderiv" vshift="-1">
+      <figure xml:id="fig_xcubedwithderiv" vshift="3">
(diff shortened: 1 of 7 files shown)


sean-fitzpatrick:
Thanks for tackling this! I'm on vacation until August 7th, but I'll look all of these over when I get back to work. 

One note: changing the chain rule title will have knock-on effects. Can you search for xrefs to the theorem? 

You might find a few that look like
`<xref ref="thm_chain_rule" text="title"/>`

Those produce a cross reference link whose text is the title of the theorem. We'll have to put "the" in front of these.

APEXCalculus:
On vacation? Don't you mean "on holiday"? No rush on looking at the changes.

I'll check out the knock-on effects of the Chain Rule title hyperlinks. Every case I saw was in the form of "the The Chain Rule." I imagine that in the original writing, I wrote "the Chain Rule". Then the latter words were replaced with the reference. But maybe somewhere the title was accounted for and an extra "the" was removed. We'll see.

sean-fitzpatrick:
I did a quick check and it looks OK. I will start merging these.

PR #368, opened 2025-08-14
Some image adjustments

sean-fitzpatrick:
A few small changes:
- One figure in Section 3.3 ran off the bottom of the page for me, so I shifted it up
- Enlarging the columns for those concavity graphs changed the page breaking in Section 3.4 (LaTeX seems to be ignoring the one that is manually inserted), and this resulted in some badly-placed figures on one page
- I noticed an example in 4.4 where the 3 column layout for the parts led to some bad line breaks. I reduced to two columns
- In 5.2 there was one image where the axis labels were colliding with other parts of the image, so I adjusted them

@APEXCalculus can you build the `latex-color-print-novideo` target and confirm that the changes to the image layout (esp. in Section 3.4) work for you? If it seems too crowded with the 4 images on Page 162 we can try a different page break.

### DIFF
--- a/ptx/sec_def_int.ptx
+++ b/ptx/sec_def_int.ptx
@@ -142,8 +142,8 @@
       \begin{tikzpicture}
       \begin{axis}[
       axis on top,
-      x label style={at={(axis description cs:0.85,0.3)},anchor=south},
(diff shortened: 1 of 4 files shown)


APEXCalculus:
Yes, I should be able to look at it tomorrow.

Issue #381, opened 2025-10-02
problem 36 in section 2.3


rhigginbottom:
Something seems wrong with problem 36 in section 2.3. ([This version of the book](https://opentext.uleth.ca/apex-calculus/sec_basic_diff_rules.html).) Sometimes it accepts the correct answer for the equation of the tangent line, and sometimes it doesn't. I've tried several versions of the problem and it seems to never accept the correct answer for the equation of the normal line.

sean-fitzpatrick:
I'll look. I think @Alex-Jordan wrote this one, because there's a lot more if-then logic than I'd manage to code into a question.

I tried 10 versions. For the first 9, both the tangent and normal lines were marked correct. For the 10th, the tangent line was correct, but the normal line was marked incorrect.

That version had f(x)=7cos(x) at pi/4.

Looking a bit closer, it seems the problem is that when the function uses cos(x), the normal line slope is coming up as 1/m instead of -1/m. But I don't see this problem when the function uses sin(x).

I'll see if I can figure out what is causing the problem.

APEXCalculus:
Thanks for looking at this so quickly, Sean.

sean-fitzpatrick:
This is fixed in #382 but I won't have time today to rebuild the HTML.

sean-fitzpatrick:
> Thanks for looking at this so quickly, Sean.

This question also happens to be on a homework assignment I set for this week, so I have extra motivation to fix it!

sean-fitzpatrick:
All versions that I host should have the updated question now. @rhigginbottom please let me know if you encounter any further issues.

rhigginbottom:
That looks great, Sean. Thanks for taking care of it! I checked a few versions that involved cos(x) and they looked like they worked. I appreciate your quick work!

sean-fitzpatrick:
Closing as completed. (Long ago. I've been delinquent in my GitHub housekeeping.)

PR #382, opened 2025-10-03
Fix sign error in cos version of normal line problem

sean-fitzpatrick:
This should fix the bug pointed out in #381.

@Alex-Jordan would you be willing to check my work before I merge this?
It is less elegant than before, but I've removed the bug, I think.

Since there was already logic to deal with the case m=0 (in which case the normal line is vertical), I just defined the slope for the normal line as -1/m. It means there are displayed answers with content like `-\left(-\frac{1}{\frac{4\sqrt{2}}{2}}\right)` but I'm willing to live with this if the correct answers are marked as such.

### DIFF
--- a/ptx/sec_deriv_basic_rules.ptx
+++ b/ptx/sec_deriv_basic_rules.ptx
@@ -1968,7 +1968,6 @@
               else {
                 $rm = Formula("1/$m");
                 $n= Formula("y=-$rm(x-pi/$a)+$k");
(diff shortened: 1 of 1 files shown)


sean-fitzpatrick:
Thanks Alex. I've checked both sine and cosine with all 4 angles and it seems to be working, so I'm going to merge.
I also removed that line, which is indeed redundant since it also appears a few lines above.

PR #411, opened 2026-01-29
temporarily remove dynamic sub randomization

sean-fitzpatrick:


sean-fitzpatrick:
There is only one question where I tried this out, and it seems to be breaking builds on Runestone, so I'm commenting out the randomization code for now.

PR #421, opened 2026-03-23
Add Washer and Shell Method exercises for volume calculations

geoff-cox:
New PROTEUS exercises for the Washer and Shell Methods

### DIFF
--- a/ptx/sec_disk.ptx
+++ b/ptx/sec_disk.ptx
@@ -1888,6 +1888,241 @@
     </solution>
   </example>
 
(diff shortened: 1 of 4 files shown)


sean-fitzpatrick:
Looks good to me. I'll merge, and rebuild on Runestone.

sean-fitzpatrick:
Ah! One gotcha: there is no `xi:include` namespace set up at the section level because we've always written the TikZ code write into the section source, rather than including it.

I'll fix that really quick.

PR #422, opened 2026-03-24
Delete exercises on Washer Method

geoff-cox:
Removed duplicate exercises related to the Washer Method from the `sec_disk.ptx` file.


sean-fitzpatrick:
Thanks. Let's blame copilot for the duplicates.

PR #425, opened 2026-05-06
Geoff's sec 10.3 PROTEUS Questions

geoff-cox:
Dumped the new PROTEUS questions just above the section exercises. Distribute them where you see fit, and just delete the ones you will not use.

### DIFF
--- a/ptx/sec_par_calc.ptx
+++ b/ptx/sec_par_calc.ptx
@@ -1093,6 +1093,282 @@
     </p>
   </subsection>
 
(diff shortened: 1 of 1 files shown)


sean-fitzpatrick:
Thanks. I've copied the ones we chose (which, for the record, are numbers 59, 62, and 64) and added them in what I think are the right places.

I did this on a proteus branch on my fork of APEX, and then I'll make a PR to move things over here.

PR #430, opened 2026-05-28
Three small content adjustments

sean-fitzpatrick:
Changes to content this time:
- In `sec_limit_def`, restrict delta to be less than 1/4 to ensure quadratic is positive.
- In `sec_sequences`, add property about limit of a continuous function of a sequence
- In `sec_limit_continuity`, add example involving continuity of a piecewise function

### DIFF
--- a/ptx/sec_limit_continuity.ptx
+++ b/ptx/sec_limit_continuity.ptx
@@ -779,7 +779,7 @@
               </md>.
               Similarly, for <m>x\gt 0</m>, <m>f(x)=5x+\cos(x^3)</m>, so
               <md>
-                \lim_{x\\to 0^+}f(x) = \lim_{x\to 0^+}(5x+\cos(x^3)) = 0+\cos(0)=1
+                \lim_{x\to 0^+}f(x) = \lim_{x\to 0^+}(5x+\cos(x^3)) = 0+\cos(0)=1
               </md>.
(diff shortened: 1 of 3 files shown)


sean-fitzpatrick:
Here's what the updated limit definition example would look like in print:

image

image

sean-fitzpatrick:
And the example on continuity at a point:

image

image

APEXCalculus:
Example 1.2.6 has a few typos in it. 
- In the line where 3/4 < x < 5/4, the line starts with an extraneous 0.

- Later, after Inequality 1.2.2 is cited, we have the correct inequality of _something_ < 29/16. When we invert, the fraction turns into 16/25, not 16/29. The incorrect 25 hangs around throughout.

Otherwise, it looks good, as does Example 1.5.10.

sean-fitzpatrick:
Fixed those mistakes, as well as one typo in the continuity example.

sean-fitzpatrick:
And one picky change: there is an `align` environment where the last line looks like `(some math) & &`
That leaves a big gap between the math and the period. I've manually placed it right after the math.

I think these should be ready, and then we just have to review the PDF layout.

sean-fitzpatrick:
I'm going to merge this so I can check that it works with all the image adjustments in the other PR.

PR #431, opened 2026-06-04
Image adjustments 2026

sean-fitzpatrick:
This ended up being a much bigger pull request than planned.

Main goal: get the margin figures lined up properly, and deal with some page fitting.

Other things that happened:
- there is now a built-in pagebreak utility, so I got rid of my own hack for this. We now specify page breaks in the publisher files.
- I got rid of a lot of the "In the following exercises" intros to exercise groups. In anticipation that we might one day want everything coded for WeBWorK, this makes sense, since the introduction gets written into every problem in the interactive version
- there were some tables in the Taylor poly and series sections that made more sense as `align*` so I made that change
- I cleared out some exercise content that doesn't look good in PDF (it's still there in HTML and WeBworK)
- I fixed some bad boxes (but not all! there are hundreds...)
- Other improvements to page breaking, especially in exercise sections

No changes to content, except for a few typo fixes.


sean-fitzpatrick:
[apex.pdf](https://github.com/user-attachments/files/28617199/apex.pdf)

@APEXCalculus here is the current PDF build. This is what we get direct from PreTeXt, with no further adjustments.
(Disclaimer: I had to submit one pull request to PreTeXt to get page breaking in exercises, so that won't work until that change is merged)

When you have time, please have a look, and please be as picky as you like about placement of figures, page breaks, etc.

I'd like to get the PDF to a point this summer where you're willing to do an APEX 5.0!

APEXCalculus:
Yes, I'll try to work on this, beginning soon. I have at least chapter 1 done so far; need to remind myself of how to edit/compile PreTeXt on my Windows machine.

sean-fitzpatrick:
Ok, great. Since I already have this pull request open, I'm also open to what I think is an easier workflow for you: 

Just go through the PDF I've built, and make a list of images that aren't quite in the right place, page breaks that seem off to you, etc. 

Post those here, and I'll make the adjustments. (You could just say things like "shift figure 3.4.5 so that the caption lines up with the text on line...")

I've done a few little tweaks already. 

One bit I was unsure of: there are a few places where a theorem or definition breaks over a page, with about 80% at the bottom of page N, and the remaining 20% at the top of page N+1.

I can add a page break before the theorem/definition, but that leaves a lot of empty space behind.

sean-fitzpatrick:
There is one pagebreak issue that I wasn't able to fix within PreTeXt (we could still manually edit the LaTeX after the fact). If the solution to an example begins with a numbered list, and it's at the bottom of the page, we can end up with the situation where the last line on a page is:

**Solution:**

And then the list begins on the next page. PreTeXt now has a way to list pagebreaks in the publisher file, but it seems like it doesn't work for example/solution.

There are about half a dozen cases of this in the entire book, which is not too bad, but if it bothers you we can manually adjust. (Two of the six instances are in the L'Hospital's rule section.)

sean-fitzpatrick:
The last commit contains some changes to make sure everything works if you manually add page breaks in the few cases where the last thing on the page is "Solution."

Fortunately there are only 5 such cases. In the publication files I've listed the xml:ids of the examples where the pagebreaks should be added, to make it easier to find the spots in the .tex file.

At this point I am going to stop working on PDF. But if you send me a list of figures you want adjusted, or pagebreaks that are needed, or places where the page needs to be enlarged, etc., I'll incorporate those changes.

APEXCalculus:
Sounds good. I'll start working on this soon, I hope.

sean-fitzpatrick:
Turns out I wasn't quite done. It occurred to me that there was a simple fix to get page breaks for the example solutions where the "Solution" title was orphaned at the bottom of a page: just put an empty paragraph at the end of the `statement` with an `xml:id`, and use that to insert the page break.

sean-fitzpatrick:
This means you shouldn't need to touch the `.tex` file at all. I think the `color-print-novideo` format is the one you're most interested in.

Make sure your pretext is up to date, and then you need to run:
- `pretext build latex-color-print-novideo`
- `cd output/latex-color-print-novideo`
- `xelatex apex.tex` (3 times)

You might be able to build the PDF in one command with `pretext build color-print-novideo` but the last time I tried this, it only ran two passes with `xelatex` instead of three. I'm running it now to see if that's still the case.

sean-fitzpatrick:
Ah! Looks like you can go straight to PDF.

First time through it'll need to build webwork and asymptote, so run `pretext build color-print-novideo`, go for lunch, and when you come back, you'll have the PDF.

sean-fitzpatrick:
@APEXCalculus I've pushed one more commit. Significant changes, but entirely organizational.
That is, this changes the targets for building the book, but not what is in the book.

Rationale:
- I wanted the names for the output to match the names for the files and folders on the server where I publish everything!
- The CLI now lets you build directly to PDF, but keep the .tex source. My hope is that most of the time, we just build the PDF, and that's it: one and done. But this way the source is there if we need to fix something.
- The primary version of APEX does not have videos, and the naming conventions should reflect that.

Main changes:
- HTML targets are now `apex-calculus` and `apex-video`. The names are self-explanatory, I think.
- PDF targets are `apex-pdf` (one-sided colour PDF, not intended for print), `apex-pdf-video` (same, but with video QR codes), `apex-print` (two-sided, black and white PDF), `apex-print-video`, `apex-color-print` (two-sided, colour PDF), and `apex-color-print-video`

There are also targets for my UofL specific books, but you shouldn't need these.

I was also using the wrong syntax for adding the Hypothes.is annotation. I fixed this for my html versions, but I left it off of the official version.

PR #442, opened 2026-07-11
2026 summer edits

APEXCalculus:
Completes vertical shift edits to Chapters 1 - 9 (through seq/series).
I shouldn't have to say it, but I've compile the .ptx code and it creates a pdf successfully, so this should be good to merge. Double-checking won't offend me. :)

### DIFF
--- a/ptx/apex.ptx
+++ b/ptx/apex.ptx
@@ -83,7 +83,7 @@
     <xi:include href="./chapter_app_of_int.ptx"/>
 
     <xi:include href="./chapter_differential_equations.ptx"/>
-
+  
     <xi:include href="./chapter_sequences_series.ptx"/>
 
     <xi:include href="./chapter_planar_curves.ptx"/>
--- a/ptx/sec_FTC.ptx
+++ b/ptx/sec_FTC.ptx
@@ -129,7 +129,7 @@
           <m>x \geq 1</m> is given by <m>A(x)=\frac12 (x)(2x)-\frac12 (1)(2)=x^2-1</m>.
         </p>
 
-        <figure xml:id="fig_ftc1b" vshift="-2">
+        <figure xml:id="fig_ftc1b" vshift="-1">
           <caption>The area of the shaded region is <m>F(x) = \int_1^x 2t\, dt</m></caption>
           <image width="47%">
             <shortdescription>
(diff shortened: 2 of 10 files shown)


sean-fitzpatrick:
Sorry for the delay. We're on summer vacation. I finally have a break from the kids this evening, so I'm testing right now.
Testing also delayed because I only brought my Windows computer with me, and there was a bug related to images in WeBWorK that only affects Windows. A fix was proposed today, so I'm testing that, too.

I am at least encouraged by the relatively small number of changes to vertical placement.

sean-fitzpatrick:
For me, the aside in the section on Integral and Comparison Tests is now pushed off the top of the page:
image

But I just realized that I built the one-sided version. I'll try two-sided. (But the pagebreaks should be the same...)

sean-fitzpatrick:
Do you also want to split up the sidebyside in Example 9.1.25? It's the same sort of thing: four sequences, with four plots, and currently the plots all come at the end. And since the sidebyside group isn't breakable, it means there's a big chunk of whitespace at the end of the text, and all the images are on the next page.

sean-fitzpatrick:
Oh, I just realized I am being dumb. I built from main, and not from this branch!! Of course things don't look right...

Sorry. I'll try again in a bit.

APEXCalculus:
No worries!

I had to delete my comment where my page numbering wasn't lining up with yours. Though I am not sure why page numbers would be off between main and this branch.

I think you'll see that the Aside to Theorem 9.3.6 does not get cut off at the top. However, it seems to appear at the bottom of the page whereas the theorem is at top. If I'm right, we should adjust this.

I had to take a break from editing as Summer School got busy. I'm now free and hope to work on this some today.

sean-fitzpatrick:
The page numbering will be off between one-sided and two-sided versions because the two-sided version will sometimes have blank pages to ensure chapters start on an odd page. 

Moving those images around will also change the number of pages. 

But page breaking is so weirdly fragile in LaTeX. There's one section (I forget which one now) where the content is identical in the two versions I make for our courses. 

One version breaks the page three lines sooner than the other.

sean-fitzpatrick:
I built the PDF (and uncovered a weird Windows encoding bug in the process...) and everything looks good!

I'll merge this, and I hope chapters 10-14 go as smoothly as this.

PR #443, opened 2026-07-28
Finished edits to Chapter 10

APEXCalculus:
This should complete all edits to chapter 10. 
Double-checking is welcome.

### DIFF
--- a/ptx/sec_polarcalc.ptx
+++ b/ptx/sec_polarcalc.ptx
@@ -126,7 +126,7 @@
                 </md>.
               </p>
 
-              <figure xml:id="fig_polcalc1" vshift="0">
+              <figure xml:id="fig_polcalc1" vshift="3">
                 <caption>The limaçon in <xref ref="ex_polcalc1"/> with its tangent line at <m>\theta=\pi/4</m> and points of vertical and horizontal tangency</caption>
             <!-- START figures/fig_polcalc1.tex -->
                 <image width="47%">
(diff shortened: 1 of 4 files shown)


ptx/sec_polarcalc.ptx
```diff
@@ -126,7 +126,7 @@
                 </md>.
               </p>
 
-              <figure xml:id="fig_polcalc1" vshift="0">
+              <figure xml:id="fig_polcalc1" vshift="3">
```
sean-fitzpatrick: Maybe the shift should be 2, not 3 here? It's the only image on the page, so it looks a bit odd to have it extend up past the top of the text:

Image

sean-fitzpatrick:
Other than my one observation above, everything here looks good. You can decide whether want to make that adjustment, and then go ahead and merge.

PS: I noticed that I must have accidentally set the wrong style file in `project.ptx` for the `apex-color-print` target: the theorems and definitions should have colour backgrounds, not black and white.

sean-fitzpatrick:
PPS: I can make that change to the project file on the pull request I have open. I haven't merged yet because there is one change that I think is a breaking change. There is a change to how we tell pretext the location of the external images folder. I think that change will break things unless I wait for the next CLI release.

Also, here is how things look if I change that vshift from 3 to 2:

image

APEXCalculus:
Took me a while to get to this. Yes, the shift looks better at 2 than 3. Made the change; merging.

PR #444, opened 2026-07-29
Technical fixes

sean-fitzpatrick:
Mostly technical stuff:

- For accessibility reasons, header rows in tables should be marked as such
- There were a few unicode characters that should have been plain text or a pretext command
- Fixes to address some code that was no longer compatible with recent changes to pretext

PR #447, opened 2026-08-12
Changes to Chapter 11

APEXCalculus:
Changes to Chapter 11 should be complete.


sean-fitzpatrick:
When I build, the figure with id `fig_quadric1` (which is Figure 11.1.26, on page 604) ends up in the wrong margin. Can you confirm this on your end?

My suggestion:  the figure is in source just above this paragraph:
```
    <p>
      We study each shape by considering <em>traces</em>,
          <idx><h>trace</h></idx>
          <idx><h>quadric surface</h><h>trace</h></idx>
      that is, intersections of each surface with a plane parallel to a coordinate plane.
      For instance,
      consider the elliptic paraboloid <m>z= x^2/4+y^2</m>,
      shown in <xref ref="fig_quadric1"/>.
      If we intersect this shape with the plane <m>z=d</m><nbsp/> (<ie/>, replace <m>z</m> with <m>d</m>),
      we have the equation:
      <md>
        <mrow>d \amp = \frac{x^2}4+y^2.</mrow>
        <intertext>Divide both sides by <m>d</m>:</intertext>
        <mrow>1 \amp = \frac{x^2}{4d} + \frac{y^2}{d}</mrow>
      </md>.
    </p>
```
I would end the paragraph after the line "shown in <xref ref="fig_quadric1"/>.", and then move the figure after that paragraph. Then start a new paragraph beginning with "If we intersect this shape..."

sean-fitzpatrick:
In the vector intro section, the last change is to change `\vnorm{0}` to `\norm{0}`. I think this should be `\norm{\vec 0}`.

We want to state that the norm of the *vector* zero is equal to the *number* 0. Right now the vector symbol is missing from the zero vector.

sean-fitzpatrick:
The last figure in 11.3 is in the wrong margin for me:
image

I'll do one more LaTeX pass to see if this changes.

sean-fitzpatrick:
Those two figures remain stubbornly in the wrong margin for me. Otherwise, things look good (aside from the missing vector symbol on the zero).

It's possible there are slight differences in our LaTeX engines? I also had to make some changes to the LaTeX XSL style file to accommodate some breaking changes in the most recent version of PreTeXt.

If you're not seeing those figures in the wrong place, could you try merging #444 into this branch to see if they move?

APEXCalculus:
I'll do some looking into this.
I've been using `pretext build apex-pdf` which is not in book format, so I haven't noticed if figures end up on the wrong margin. Compiling now - sorry, should have known better.

And the `\norm{\vec 0}` is right. I remember making the edit; while I forget the details, I recall that `\vnorm{0}` looked/was wrong.

sean-fitzpatrick:
I've been using `pretext build apex-color-print`.

APEXCalculus:
Yes - am currently compiling the .tex from that build a second time and will look at the figures shortly.

sean-fitzpatrick:
One note: if you're building the color print version, you'll want to compare page breaks in publication-pdf to publication-colour-print.

APEXCalculus:
I think I have my system set up correctly now and I can get stable builds of `apex-color-print`. 
This is hard to run down, but it seems to me that the figures in Figure 11.3.6 are sized differently for `apex-pdf` and `apex-color-print`, with the latter being just a shade smaller, meaning the page break comes in a different place.
Is this something you can verify? And figure out why it is happening?
(PS - going awol for a while as I have to run errands. Will get back to this asap.)

sean-fitzpatrick:
I'll build both versions today and compare. LaTeX can be infuriating sometimes. For the versions I build for University of Lethbridge, there is one chapter where the content of the first page is *identical* in both versions, and yet the page break is off by an entire paragraph!

I'll look to see if there is some subtle difference in the styling used in the two versions.

sean-fitzpatrick:
I see one difference in the publication files that could be relevant.
I'm going to do some reconciliation across the different publication file versions to make sure everything lines up correctly.

I'll push this to the Chapter 11 branch, so you'll want to make sure you pull before making any further changes.

sean-fitzpatrick:
OK: I mentioned how LaTeX can be infuriating. On the very first page of Chapter 11, the `apex-pdf` and `apex-color-print` versions have different page breaks.

apex-pdf:
image

apex-color-print:
image

I ran a diff on the `.tex` files for the two versions. For the first three pages of the chapter, they are *identical*. Why the page breaks are different is a complete mystery. That said, Section 11.1 comes out OK despite the initial discrepancy.

sean-fitzpatrick:
11.2 is fine. In 11.3 we run into one badly-placed figure:
image

I am going to push a change that adds a page break before this example, and lowers the figure a bit so it doesn't run off the top of the next page.

sean-fitzpatrick:
Here's the result with a page break. Not sure if you want to shift the figure up by half a centimetre or leave it as is:
image

sean-fitzpatrick:
I found that Figures 11.3.10 and 11.3.26 weren't placed well: the first was too high at the top of the page, and the second was too low. This was for the `apex-color-print` target.

If the adjustments I just made are acceptable to you, you can merge this branch.

APEXCalculus:
Your final changes looked good ... and go figure I had un-pushed commits on my local branch that somewhat conflicted with yours. Got them resolved, then lowered 11.3.10 one more half cm.

sean-fitzpatrick:
Ok, great! Does that take us to the end, then? 
(As long as we don't look too carefully at the appendices)

APEXCalculus:
Just about!
I'd like to work on the Answers section a bit as I get the time.
It seems silly/terrible to say this, but I *finally* got a productive
workflow in Codespaces.
For a while I had to manually edit the .tex file in order for it to
compile; once that was fixed, Codespace still wouldn't actually compile the
.tex without error; now that is fixed and I can make edits to the .ptx, run
one command, and (a few minutes later) see the result.

PR #449, opened 2026-08-13
2026 summer edits chpt12

APEXCalculus:
This is mostly just vertical figure adjustment, but there are a few other changes.

Definition 12.2.8 had 2 asides that accompanied it. I have merged these into one aside. 

Within the Solution to Exercise 12.4.10, there was an aside that referred to a "flaw" in the definition of the unit normal vector `N` wherein `T'` could not be `\vec 0`; however, lines have this property! I moved the aside into text that follows Theorem 12.4.12 where it seems to fit better. I also changed the wording a little.

### DIFF
--- a/ptx/sec_curvature.ptx
+++ b/ptx/sec_curvature.ptx
@@ -208,7 +208,7 @@
           We find it with <m>\vec r(2) = \la 1/5, 18/5\ra</m>.
         </p>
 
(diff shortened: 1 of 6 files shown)


sean-fitzpatrick:
Everything in this chapter looks good to me.

PR #464, opened 2026-09-11
Fixes simple typos in the text ... (and hopefully changing an "l" to …

APEXCalculus:
…an "r" doesn't cascade and cause images to move)

Should be simple to merge; I've compiled and it looks right.


sean-fitzpatrick:
Looks good to me!

